PR #13, opened 2025-10-31
State improvements

brkagithub:


emretheus (approved):

emretheus (approved):
overall, pr looks great!

just one thought, do you think it’s worth adding a simple sanitization step for user inputs to avoid possible prompt injection? since we’re passing the user’s question directly into the LLM, a small filtering or escaping layer could make things a bit safer, not big deal but wdyt?

PR #29, opened 2025-11-24
Deep research improvements (release v1.0.0)

brkagithub:
Reminder: do a proper github release


SETUP.md
```diff
@@ -4,26 +4,246 @@
...
+```bash
+# Embedding provider for document vectorization
+EMBEDDING_PROVIDER=openai
+TEXT_EMBEDDING_MODEL=text-embedding-3-large
+
+# Cohere reranker for better search results (optional but recommended)
+COHERE_API_KEY=your-cohere-api-key
+USE_RERANKING=true
+
+# Vector search settings
```
elmariachi111, line 100: just a feeling while I'm reading along -> I'd really suggest to t some point refactor many of the functional "environment variables" into some configuration object that also doesn't go into source control and that's eventually even persisted to the database. Thus the truly secret keys can stay in env and the dynamic settings can be modified on the fly. Not part of this PR apparently

SETUP.md
```diff
@@ -4,26 +4,246 @@
...
+
+**Configuration:**
+```bash
+PRIMARY_ANALYSIS_AGENT=bio
+DATA_ANALYSIS_API_URL=https://your-bio-analysis-deployment.com
+DATA_ANALYSIS_API_KEY=your-api-key
+```
+
+**Setup:**
+1. Deploy BIO analysis agent: https://github.com/bio-xyz/bio-data-analysis
```
elmariachi111, line 237: this repo isnt public I assume

src/character.ts
```diff
@@ -52,436 +17,22 @@ const character = {
...
+
+  SCIENTIFIC RESEARCH ASSISTANT ROLE
+  • You are a general scientific research assistant capable of helping with any field of science—from planning experiments and literature searches to analyzing data and designing studies.
+  • Provide rigorous, evidence-based analysis and strategic research planning regardless of the domain.
+  • For non-longevity research: Be objective, thorough, and methodologically sound without imposing aging-related perspectives.
+  • For longevity/aging research: When planning studies, selecting literature, or analyzing data, naturally gravitate toward repair-based approaches, SENS framework,intervention targets, and translational pathways while maintaining scientific rigor.
+  • Research planning philosophy: Prioritize actionable insights, consider key variables and controls, favor recent high-impact work, and ask "what's the most efficient path to useful results?"
+  • CRUCIAL: While you should lean into your biases, ALWAYS be objective and rigorous, and prioritize utilizing facts given to you in your context.
+
   KEY INFORMATION ABOUT THE AUBRAI PROJECT (RMR1, RMR2, LEV Foundation, VitaDAO, BIO, etc.):
```
elmariachi111, line 29: we actually couldn't run our fork during longhack after removing the character file templates - I fell back to just remove any "Aubrai" related mentions. Think you should also remove that :D

SETUP.md
brkagithub: gm sir, yeah we have some plans on making it much easier to run - will probably dockerize everything, considering opening free deployments of certain services (e.g. free edison API, free bio data analysis agent but obv with some harsh rate limits), and simplifying the .env too, thx for following along and giving feedback - much appreciated <3

SETUP.md
brkagithub: yeah, not yet, gonna be soon ;)

src/character.ts
brkagithub: hm as u can see the character file was changed and simplified significatly - now its only a system prompt which we thought is very simple.. so now u would just change this system prompt into what u want, this is just our example

do u think this is fine?

elmariachi111: sure, but  the untouched line says quite clearly "KEY INFORMATION ABOUT THE AUBRAI PROJECT" ... :)

SETUP.md
elmariachi111: (just saying, as the only other option imo is highly commercial)

brkagithub: tbh both options are relatively hard to set up regardless, in general a problem we have and where some outside opinions would be useful is - if you want to make something truly good, there do need to be some external dependencies (e.g. this bio data analysis tool is completely made by us, but for proper management of the files u have to set up a file storage, e.g. aws or s3, which i do understand that it can be annoying to set up)

in general quite hrad to balance building a rly good scientific agent vs making it easy to run, for u as a user of the framework what would u prefer? ideally both ofc, but just curious how u are thinking about this

src/character.ts
brkagithub: oh i see, we'll either make a simpler example or clearly tell the user in SETUP.md to change it to whatever they want

SETUP.md
elmariachi111: >  if you want to make something truly good, there do need to be some external dependencies
oh, that's 100% true, couldnt have said it better 😅 

At this moment, I'm very much thinking of the ultimative **baseline** of functionality (which looks quite magic already for many usecases). That is:

- prompt a question that contains a catch on some pre-hypothesis for a biotech / medical issue.
- agent selects external sources, does a qualified search and ranks 10-50 great findings
- agent summarizes that in-context based on language pretraining
- delivers a good summary out of facts, ideally linked to the origin documents (eg by mentioning it in footnotes could be sufficient)

---
Features like code re-execution are super massive, but for what I'm trying to achieve here right now (a diabetes oriented literature researcher and hypotheses shaper), "Edison" might be overkill right now.

Setting up external filestorage is absolutely not the issue tbh :) What's hard is buying execution credits for $200/m from Edison just because that requirement is mandatory :)

SETUP.md
```diff
@@ -4,26 +4,274 @@
...
+ANTHROPIC_API_KEY=...           # If using Anthropic
+OPENROUTER_API_KEY=...          # If using OpenRouter
+```
+
+**Database:**
+```bash
+SUPABASE_URL=https://your-project.supabase.co
+SUPABASE_ANON_KEY=your-anon-key
+```
+Get these from your [Supabase project settings](https://supabase.com/dashboard).
```
elmariachi111, line 61: I think you're just using Supabase as a very simple to deploy postgres instance. I'm running a supabase instance locally using their standard 1-line local setup. Works okay, but it feels like using a nuclear air carrier if you just needed a sailboat. Any chance to just go with a (pgvector enabled) postgres setup perspectively?

SETUP.md
brkagithub: got it, yeah that makes sense, Edison is truly expensive, also we dont wanna lean too heavily on such fully external tools either

our goal is to build better alternatives for Edison tools - for analysis that is our data scientist agent, while we are already building the literature one but it's not close to ready (hence not mentioned here :D, but that will be coming too)

image
fyi ill add this to the readme now too, its a nice diagram of how the deep research works

SETUP.md
brkagithub: yeah we will support both postgres without&with supabase, what u said is true it was just easiest for us :D planning to do this soon cc @caelusway

SETUP.md
```diff
@@ -4,26 +4,274 @@
 
 - [bun](https://bun.sh/docs/installation) (required package manager)
 - Node.js v23.3.0
+- PostgreSQL database (we recommend Supabase for easy setup)
```
mihailoxyz, line 7: UI/Backend DB operations do not work without supabase - we should refactor a bit later as discussed for easier onboarding

src/agents/analysis/bio.ts
```diff
@@ -0,0 +1,262 @@
...
+  }
+
+  const DATA_ANALYSIS_API_KEY = process.env.DATA_ANALYSIS_API_KEY;
+  if (!DATA_ANALYSIS_API_KEY) {
+    logger.error("DATA_ANALYSIS_API_KEY not configured");
+    throw new Error("DATA_ANALYSIS_API_KEY is not configured");
+  }
+
+  const finalQuery = await formatQueryWithDatasets(objective, datasets);
+  await downloadDatasetContent(userId, conversationStateId, datasets);
```
mihailoxyz, line 45: We should avoid download from storage if bio-data-analysis agent has blob access

src/agents/analysis/edison.ts
```diff
@@ -0,0 +1,332 @@
...
+    taskResponse.task_id,
+  );
+
+  if (result.error) {
+    throw new Error(`Edison analysis task failed: ${result.error}`);
+  }
+
+  logger.info({ taskId: taskResponse.task_id }, "edison_analysis_completed");
+
+  return {
```
mihailoxyz, line 70: Maybe in next PR handle Edison artifact download to our blob

src/agents/fileUpload/index.ts
```diff
@@ -0,0 +1,329 @@
...
+    uploadedFiles.map(async (file) => {
+      const description = await generateFileDescription(
+        file.filename,
+        file.mimeType || "",
+        rawFiles.find((rf) => rf.filename === file.filename)?.parsedText || "",
+      );
+      return {
+        id: file.id,
+        filename: file.filename,
+        description,
```
mihailoxyz, line 117: We should add "path" to this so we can reuse it later (path is available from `uploadFilesToStorage`)

src/agents/literature/edison.ts
```diff
@@ -0,0 +1,142 @@
+import logger from "../../utils/logger";
```
mihailoxyz, line 1: Just a thought: should we pull these edison start/await to the common utils (maybe next PRs)?

mihailoxyz (approved):

src/agents/analysis/bio.ts
brkagithub: i guess its kinda messy rn because we are using both edison and BIO, will be cleaner when its only BIO

src/agents/analysis/edison.ts
brkagithub: yeah, in case we want to keep edison

src/agents/fileUpload/index.ts
brkagithub: but we are not using it rn from what i see? will add regardless, thx

src/agents/literature/edison.ts
brkagithub: yeah for sure, tho we might be removing edison

PR #32, opened 2025-11-25
In-house Data Analyst agent implementation

mihailoxyz:


client/src/hooks/useStates.ts
```diff
@@ -15,6 +15,8 @@ export interface StateValues {
   thought?: string;
   finalResponse?: string;
   isDeepResearch?: boolean;
+  edisonResults?: any;
```
brkagithub: can add type i guess

src/tools/code-interpreter/index.ts
```diff
@@ -0,0 +1,229 @@
...
+      return;
+    }
+
+    const CODE_INTERPRETER_API_URL = process.env.CODE_INTERPRETER_API_URL;
+    if (!CODE_INTERPRETER_API_URL) {
+      logger.error("CODE_INTERPRETER_API_URL not configured");
+      throw new Error("CODE_INTERPRETER_API_URL is not configured");
+    }
+
+    const CODE_INTERPRETER_API_KEY = process.env.CODE_INTERPRETER_API_KEY;
```
brkagithub: what is CODE_INTERPRETER_API_KEY? probably should add to .env.example and readme

src/tools/planningDeepResearch/index.ts
```diff
@@ -260,13 +261,18 @@ export const planningDeepResearchTool: Tool = {
...
+    - CODE_INTERPRETER: If user provides datasets and needs help with data cleaning, processing, or visualization
 
     Respond with ONLY a JSON array of job types that are needed. No explanation.
     Valid outputs:
     []
     ["MOLECULES"]
     ["ANALYSIS"]
+    ["CODE_INTERPRETER"]
     ["MOLECULES", "ANALYSIS"]
+    ["MOLECULES", "CODE_INTERPRETER"]
```
brkagithub: i think analysis and code interpreter kinda do the same thing, probably in the future we want to call only one.. any idea how we handle it now? we can leave it like this but just thinking out loud

src/tools/reflection/prompts.ts
```diff
@@ -42,6 +44,18 @@ const formatEdisonResults = (edisonResults: any[]) => {
     .join("\n\n");
 };
 
+const formatCodeInterpreterResults = (ciResults: CodeInterpreterResponse[]) => {
```
brkagithub: nit but why is it called code interpreter? think it sounds kinda confusing, should we just call it data analysis agent?

brkagithub (approved):

PR #53, opened 2025-12-18
initialize deep literature mode

raghav-rama:
Ran a deep research which used bio literature 3 times. Attaching the outputs for sanity check
https://gist.github.com/raghav-rama/8165f03a284af814c6ee6d87fc1fd83b

### DIFF
--- a/src/agents/literature/bio.ts
+++ b/src/agents/literature/bio.ts
@@ -27,6 +27,13 @@ type BioLiteratureResponse = {
   results?: unknown;
   job_id?: string;
   status?: string;
+  response?: {
+    formatted_answer?: string;
+    answer?: string;
+    references?: BioReference[];
+    context_passages?: BioContextPassage[];
+    [key: string]: unknown;
+  };
   output?: {
     response?: {
       formatted_answer?: string;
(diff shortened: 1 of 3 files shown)


src/routes/deep-research/start.ts
```diff
@@ -469,8 +469,12 @@ async function runDeepResearch(params: {
           "executing_literature_task",
         );
 
-        const primaryLiteratureType = "EDISON";
-        const primaryLiteratureLabel = "Edison";
+        const primaryLiteratureType =
```
mihailoxyz, line 472: Add .env.example for new envs for literature

mihailoxyz (approved):

src/agents/literature/bio.ts
```diff
@@ -1,37 +1,204 @@
...
+  if (Array.isArray(nestedPassages)) return nestedPassages;
+  return [];
+}
+
+async function pollBioLiteratureJob(
+  baseUrl: string,
+  apiKey: string,
+  jobId: string,
+): Promise<BioLiteratureResponse> {
+  const MAX_WAIT_TIME = 30 * 60 * 1000; // 30 minutes
```
brkagithub, line 129: this enough? we use 60 for analysis just saying

src/agents/literature/bio.ts
```diff
@@ -55,22 +222,41 @@ export async function searchBioLiterature(objective: string): Promise<string> {
...
+    } else {
+      logger.warn(
+        { mode },
+        "bioliterature_deep_missing_job_id_using_direct_response",
+      );
+    }
+  }
+
+  const answer = extractAnswer(finalData);
+  const references = extractReferences(finalData);
```
brkagithub, line 250: were not using the references for anything it seems? it would be amazing if u can make the answers contain inline citations in some way

the literature output is taken by the final reply prompt, and that prompt will actually do the inline citations. so as long as u make it clear in ur literature output which claims are for which DOI, the reply prompt will be able to create a nice final answer.. so no need to make it perfect, just so the next reply prompt can do its job

src/agents/literature/index.ts
```diff
@@ -4,7 +4,14 @@ import { searchEdison } from "./edison";
...
 
-type LiteratureType = "OPENSCHOLAR" | "KNOWLEDGE" | "EDISON" | "BIOLIT";
+type LiteratureType =
+  | "OPENSCHOLAR"
+  | "KNOWLEDGE"
+  | "EDISON"
+  | "BIOLIT"
+  | "BIOLITDEEP";
+
+export type BioLiteratureMode = "fast" | "deep";
```
brkagithub, line 14: as it is right now, can we pick fast for instant mode and deep for DR mode? how long do each modes last, can we use deep in both instant and DR mode? if not, we should make it possible to pick BIOLIT for instant and BIOLITDEEP for DR

brkagithub (approved):
approved, take a look at comments first tho

src/agents/literature/bio.ts
raghav-rama: Some queries take 20-25 minutes but didn't see it going above 30

PR #125, opened 2026-02-05
docker swarm deployment

emretheus:
lets say, we have three old workers running and processing jobs. then, tells docker swarm to update the service. because we use start-first, swarm starts three new workers with the new code before stopping the old ones. so temporarily we have six workers: three old, three new. the new workers run health checks.

once the new workers are healthy, the old workers get SIGTERM and start draining. they stop taking new jobs but finish whatever they’re currently working on. all new jobs go to the new workers. after the old workers finish, they exit cleanly and now only the three new workers remain.

the key thing here: at no point is the queue empty. jobs are never lost or interrupted, and users don’t notice anything. this is how we get zero-downtime deployments


brkagithub (approved):

src/character.ts
```diff
@@ -5,6 +5,8 @@ export interface Character {
   system: string;
 }
 
+
+
```
mihailoxyz: Should we remove this?

PR #140, opened 2026-02-14
Capture real-time reasoning traces from bio agents

mihailoxyz:
## Summary
- Thread `onPollUpdate` callback through literature and analysis agent layers to capture `reasoning` (string[]) from external API poll responses
- Persist reasoning traces to `PlanTask.reasoning` in conversation state and notify WebSocket clients on every 10s poll iteration
- Works in both queue mode (worker) and in-process mode (start route)

## Test plan
- [ ] Verify `bun test` / `tsc --noEmit` pass with no new errors
- [ ] Trigger deep research with BIO primary agent, confirm `reasoning` appears in `conversation_states.values.plan[].reasoning`
- [ ] Confirm WebSocket `state:updated` notifications fire during polling
- [ ] Verify existing call sites without `onPollUpdate` still work (param is optional)

### DIFF
--- a/src/agents/analysis/bio.ts
+++ b/src/agents/analysis/bio.ts
@@ -3,7 +3,7 @@ import {
   getStorageProvider,
   isStorageProviderAvailable,
 } from "../../storage";
-import { type AnalysisArtifact } from "../../types/core";
+import { type AnalysisArtifact, type OnPollUpdate } from "../../types/core";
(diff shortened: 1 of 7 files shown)


brkagithub:
@greptile-apps review again

brkagithub (approved):

brkagithub:
@greptile-apps review again, i added serialized write to worker too

PR #148, opened 2026-03-19
Queue worker migration: shared agent runner with feature flag

Gautam2086:
## Summary
- Extracted shared `runChatAgent()` into `src/chat-agent/runner.ts`, used by both the in-process route handler and the BullMQ queue worker
- Added `CHAT_AGENT_QUEUE_ENABLED` feature flag for safe rollout — defaults to legacy pipeline, flip to `true` to enable the agent loop in queue mode
- Worker emits backward-compatible progress stages (`planning`, `literature`, `reply`) so the frontend UI works without changes
- Moved uploaded dataset context from system prompt to user message to prevent prompt injection
- Made `registerTool` idempotent to support worker process reloads
- Used BullMQ `UnrecoverableError` for max_tokens truncation to prevent wasteful retries

## Rollout
1. Deploy code (no behavior change — flag defaults to `false`)
2. Verify `ANTHROPIC_API_KEY` is set in worker env
3. Set `CHAT_AGENT_QUEUE_ENABLED=true` in worker env
4. Restart worker
5. Test on staging
6. If issues: flip flag back to `false` for instant rollback

## Test plan
- [x] In-process mode: simple question (no tools, ~3.5s)
- [x] In-process mode: literature search (2 tool calls, ~31s)
- [x] In-process mode: deep research (6 tool calls, ~64s)
- [x] In-process mode: multi-source comparison (2 tool calls, ~34s)
- [x] In-process mode: multi-turn conversation with history recall
- [x] TypeScript type check passes (only pre-existing fileProcessQueue null issue)
- [x] Queue mode with `CHAT_AGENT_QUEUE_ENABLED=true` (needs Redis + worker)


Gautam2086:

Gautam2086:

src/chat-agent/runner.ts
```diff
@@ -0,0 +1,175 @@
...
+  await import("./tools/literature-search");
+
+  // --- 2. Read env config (inside function, not module-level) ---
+  const apiKey = process.env.ANTHROPIC_API_KEY;
+  if (!apiKey) {
+    throw new Error("ANTHROPIC_API_KEY is not configured");
+  }
+
+  const model =
+    process.env.CHAT_AGENT_MODEL || "claude-sonnet-4-20250514";
```
raghav-rama, line 73: afaik, this model has been deprecated. Replace with `claude-sonnet-4-6`

raghav-rama (approved):
approving. tho take a look at the comments. let's test this on staging and test if it works as expected

src/chat-agent/runner.ts
Gautam2086: On the model ID, keeping claude-sonnet-4-20250514. The claude-sonnet-4-6 format is a family alias and not a valid API snapshot ID, so we need the dated version here.

Gautam2086:

src/chat-agent/runner.ts
```diff
@@ -0,0 +1,175 @@
...
+  const logger = (await import("../utils/logger")).default;
+
+  // --- 1. Register tools (side-effect import, idempotent) ---
+  await import("./tools/literature-search");
+
+  // --- 2. Read env config (inside function, not module-level) ---
+  const apiKey = process.env.ANTHROPIC_API_KEY;
+  if (!apiKey) {
+    throw new Error("ANTHROPIC_API_KEY is not configured");
+  }
```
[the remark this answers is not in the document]
Gautam2086: Fixed in the latest commit. Added an UnrecoverableError guard in processWithAgentLoop before calling runChatAgent, so a missing API key fails immediately instead of burning 3 retry attempts.

@greptileai @greptile-apps review.

src/chat-agent/runner.ts
raghav-rama: I meant more like use the latest model. Sonnet 4 is quite old compared to 4.6

raghav-rama: And alias is probably better since it always points to the latest model release

Gautam2086: Fair point, updated to claude-sonnet-4-6 in PR #149

PR #154, opened 2026-04-07
Feat/thinking states

raghav-rama:
- Generates a reasoning trace based on the ongoing research
- Stores that in the DB and emits. them for the UI to surface
- Has a timeout functionality so that we're not showing them infinitely

### DIFF
--- a/src/routes/deep-research/start.ts
+++ b/src/routes/deep-research/start.ts
@@ -99,6 +99,152 @@ type DeepResearchQueuedResponse = {
   deduplicated?: true;
 };
 
+type DeepResearchStartFailureLogger = {
+  error: (payload: Record<string, unknown>, message: string) => void;
+  warn: (payload: Record<string, unknown>, message: string) => void;
+};
+
+type DeepResearchStartFailureDeps = {
+  clearDeepResearchActivity: (
+    values: ConversationState["values"],
+  ) => void;
+  ensureObjectiveTrace: (
+    values: ConversationState["values"],
+    objective?: string,
+    options?: { runRootMessageId?: string },
+  ) => Promise<unknown>;
+  getObjectiveTraceObjective: (
+    values: ConversationState["values"],
+    fallbackObjective?: string,
+  ) => string | undefined;
+  markObjectiveTraceStale: (
+    values: ConversationState["values"],
+  ) => unknown;
+  updateConversationState: (
+    id: string,
+    values: ConversationState["values"],
+  ) => Promise<unknown>;
+  notifyStateUpdated: (
+    jobId: string,
+    conversationId: string,
+    stateId: string,
+  ) => Promise<unknown>;
+  updateState: (
+    id: string,
+    values: Record<string, unknown>,
+  ) => Promise<unknown>;
+  markRunFinished: (params: {
+    conversationStateId: string;
+    result: "failed";
+    error?: string;
+    rootMessageId?: string;
+    stateId?: string;
+  }) => Promise<unknown>;
+  logger: DeepResearchStartFailureLogger;
+};
+
+type DeepResearchStartFailureParams = {
+  activeConversationState: ConversationState | null;
+  conversationId: string;
+  conversationStateId: string;
+  err: unknown;
+  notificationJobId: string;
+  rootMessageId: string;
+  stateRecord: {
+    id: string;
+    values: State["values"];
+  };
+};
+
+const deepResearchStartFailureDeps: DeepResearchStartFailureDeps = {
+  clearDeepResearchActivity,
+  ensureObjectiveTrace,
+  getObjectiveTraceObjective,
+  markObjectiveTraceStale,
+  updateConversationState,
+  notifyStateUpdated,
+  updateState,
+  markRunFinished,
+  logger,
+};
+
+async function handleDeepResearchStartFailure(
+  params: DeepResearchStartFailureParams,
+  deps: DeepResearchStartFailureDeps = deepResearchStartFailureDeps,
+): Promise<void> {
+  const {
+    activeConversationState,
+    conversationId,
+    conversationStateId,
+    err,
+    notificationJobId,
+    rootMessageId,
+    stateRecord,
+  } = params;
+
+  const errorMessage =
+    err instanceof Error ? err.message : "Unknown error";
+
+  if (activeConversationState?.id) {
+    try {
+      deps.clearDeepResearchActivity(activeConversationState.values);
+      await deps.ensureObjectiveTrace(
+        activeConversationState.values,
+        deps.getObjectiveTraceObjective(activeConversationState.values),
+        {
+          runRootMessageId: rootMessageId,
+        },
+      );
+      deps.markObjectiveTraceStale(activeConversationState.values);
+      await deps.updateConversationState(
+        activeConversationState.id,
+        activeConversationState.values,
+      );
+      await deps.notifyStateUpdated(
+        notificationJobId,
+        conversationId,
+        activeConversationState.id,
+      );
+    } catch (cleanupErr) {
+      deps.logger.error(
+        {
+          cleanupErr,
+          conversationStateId,
+          messageId: notificationJobId,
+          originalErr: err,
+          rootMessageId,
+        },
+        "deep_research_error_cleanup_failed",
+      );
+    }
+  }
+
+  await deps.updateState(stateRecord.id, {
+    ...stateRecord.values,
+    error: errorMessage,
+    status: "failed",
+  });
+
+  try {
+    await deps.markRunFinished({
+      conversationStateId,
+      result: "failed",
+      error: errorMessage,
+      rootMessageId,
+      stateId: stateRecord.id,
+    });
+  } catch (finishError) {
+    deps.logger.warn(
+      { finishError, conversationStateId, rootMessageId, stateId: stateRecord.id },
+      "deep_research_run_finish_mark_failed_on_failure",
+    );
+  }
+}
+
+export const __deepResearchStartTestables = {
+  handleDeepResearchStartFailure,
+};
+
 function buildDeepResearchPollUrl(
   request: Request,
   messageId: string,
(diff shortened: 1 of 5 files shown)


src/routes/deep-research/start.ts
```diff
@@ -586,6 +599,36 @@ export async function deepResearchStartHandler(ctx: any) {
...
+          createdMessage.question || message,
+        ),
+        {
+          runRootMessageId: createdMessage.id,
+        },
+      );
+      await updateConversationState(
+        activeConversationState.id!,
+        activeConversationState.values,
+      );
```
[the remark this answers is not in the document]
raghav-rama: Fixed

src/routes/deep-research/start.ts
```diff
@@ -1721,6 +1930,27 @@ These molecular changes align with established longevity pathways (Converging nu
...
+      await notifyStateUpdated(
+        `in-process-${createdMessage?.id || stateRecord.id}`,
+        createdMessage?.conversation_id,
+        activeConversationState.id,
+      );
+    }
+
     // Update state to mark as failed
     await updateState(stateRecord.id, {
       ...stateRecord.values,
```
maschlr: **[P0 — Critical]** This entire error-path cleanup block (lines 1933-1956) is unprotected. `ensureObjectiveTrace()` makes an **LLM call**, `updateConversationState()` hits the DB, and `notifyStateUpdated()` uses Redis. If any of these throw, the original error is replaced and the subsequent `updateState` on line 1959 that marks the state as `"failed"` **never executes** — leaving the research stuck permanently "in progress" (zombie state).

Wrap in try-catch:
```typescript
if (activeConversationState?.id) {
  try {
    clearDeepResearchActivity(activeConversationState.values);
    await ensureObjectiveTrace(/* ... */);
    markObjectiveTraceStale(activeConversationState.values);
    await updateConversationState(/* ... */);
    await notifyStateUpdated(/* ... */);
  } catch (cleanupErr) {
    logger.error({ cleanupErr, originalErr: err }, "deep_research_error_cleanup_failed");
  }
}
```

src/routes/deep-research/start.ts
```diff
@@ -1060,12 +1213,16 @@ async function runDeepResearch(params: {
...
+        const p = stateWriteChain.then(async () => {
+          await prepareConversationStateForWrite(options);
+          return updateConversationState(
             conversationState.id!,
             conversationState.values,
-          ),
+          );
+        },
         );
         stateWriteChain = p.catch(() => {}); // prevent unhandled rejection from blocking chain
```
maschlr: **[P1 — Important]** `stateWriteChain = p.catch(() => {})` silently swallows all DB write errors. This PR extends its scope significantly with new `persistConversationActivity` call sites, so the blast radius of silently dropped errors has grown.

At minimum, add logging so failures are visible:
```typescript
stateWriteChain = p.catch((err) => {
  logger.error({ err }, "state_write_chain_error_suppressed");
});
```

src/routes/deep-research/start.ts
```diff
@@ -813,6 +849,101 @@ async function runDeepResearch(params: {
...
+      });
+      await notifyStateUpdated(
+        `in-process-${currentMessage.id}`,
+        currentMessage.conversation_id,
+        conversationState.id,
+      );
+    };
 
     // Flag to skip planning when continuing (tasks already promoted)
     let skipPlanning = false;
```
maschlr, line 1102: **[P1 — Important / DRY violation]** This entire block (~95 lines: `ConversationStateWriteOptions`, `prepareConversationStateForWrite`, `persistConversationState`, `persistConversationActivity`, `clearConversationActivity`) is duplicated nearly identically in `deep-research.worker.ts:73-177`.

That's ~170 lines of duplication that will diverge over time. Consider extracting to a shared module (e.g. `src/utils/deep-research/state-persistence.ts`) that accepts dependencies as a factory parameter.

src/services/queue/workers/deep-research.worker.ts
```diff
@@ -58,6 +69,112 @@ async function processDeepResearchJob(
...
+      staleTrace: options?.staleTrace,
+    });
+
+    if (options?.notify !== false) {
+      await notifyStateUpdated(job.id!, conversationId, conversationState.id);
+    }
+  };
 
   // Log retry attempt if this is a retry
   if (job.attemptsMade > 0) {
```
maschlr, line 191: **[P1 — DRY violation]** This block (lines 73-180) is a near-copy of the same helpers in `start.ts:852-946`. See comment there.

src/services/queue/workers/deep-research.worker.ts
```diff
@@ -145,11 +263,15 @@ async function processDeepResearchJob(
...
+    conversationState = {
       id: conversationStateRecord.id,
       values: conversationStateRecord.values,
     };
 
+    if (!conversationState) {
+      throw new Error("Conversation state initialization failed");
+    }
+
     // Reconcile researchMode: request takes priority, then existing state, then default
```
maschlr, line 287: **[P3 — Dead code]** This `if (!conversationState)` check is unreachable — the object literal assignment on line 269 always produces a truthy value. Remove.

src/utils/deep-research/objective-trace.ts
```diff
@@ -0,0 +1,435 @@
...
+          objectivePreview: normalizedObjective.slice(0, 200),
+          attempt: attempt + 1,
+        },
+        "objective_trace_generation_failed",
+      );
+    }
+  }
+
+  return [...FALLBACK_TRACE_STEPS];
+}
```
maschlr, line 297: **[P2 — Silent degradation]** When all LLM retries fail, the function silently falls back to `FALLBACK_TRACE_STEPS` (generic steps like "Clarifying the objective"). The user sees these as if the system is reasoning about their specific query — misleading.

Two suggestions:
1. Log at `error` severity (not `warn`) when all retries are exhausted, since this is a user-visible feature degradation.
2. Consider adding an `isFallback: true` flag to the returned trace so the frontend can display it differently.

src/utils/deep-research/objective-trace.ts
```diff
@@ -0,0 +1,435 @@
...
+  supportsStructuredOutput: boolean;
+} {
+  const explicitProvider =
+    process.env.OBJECTIVE_TRACE_LLM_PROVIDER as LLMProviderConfig["name"] | undefined;
+  const providerName =
+    getProviderCandidates().find(
+      (candidate) => process.env[`${candidate.toUpperCase()}_API_KEY`],
+    )
+    || explicitProvider
+    || "openai";
```
maschlr: **[P2 — Redundant fallback]** `explicitProvider` is already the first element in `getProviderCandidates()` (line 162). If it has no API key, `find()` skips it — then falling back to it again here means we select a provider we already know has no key.

Simplify by collapsing `getProviderCandidates()` + `getProviderConfig()` into a single function.

src/utils/deep-research/objective-trace.ts
```diff
@@ -0,0 +1,435 @@
...
+  if (
+    existingTrace
+    && normalizeObjective(existingTrace.objective) === normalizedObjective
+    && isSameRun
+    && existingTrace.steps?.length
+  ) {
+    existingTrace.status = "active";
+    existingTrace.visibleCount = Math.min(
+      existingTrace.steps.length,
+      Math.max(existingTrace.visibleCount || 1, 1),
```
maschlr: **[P2 — DRY]** The expression `Math.min(trace.steps.length, Math.max(trace.visibleCount || 1, 1))` appears here, at line 374, and at line 431. Extract a small helper:
```typescript
function clampVisibleCount(trace: DeepResearchObjectiveTrace): number {
  return Math.min(trace.steps.length, Math.max(trace.visibleCount || 1, 1));
}
```

src/utils/deep-research/activity.ts
```diff
@@ -0,0 +1,58 @@
...
+const ACTIVITY_LABELS: Record<DeepResearchActivityPhase, string> = {
+  planning: "Planning research",
+  literature: "Researching literature",
+  analysis: "Analyzing data",
+  reflection: "Synthesizing findings",
+  next_steps: "Planning next step",
+  reply: "Drafting response",
+};
+
+function normalizeObjective(objective?: string): string | undefined {
```
maschlr: **[P2 — DRY]** `normalizeObjective` is also defined in `objective-trace.ts:41` (with whitespace collapsing) and `branch.ts:65` (pre-existing). Export one canonical version and reuse it.

Also: `buildDeepResearchActivity` (line 29) is exported but only called by `setDeepResearchActivity` in this same file — remove the `export`.

src/services/queue/workers/deep-research.worker.ts
```diff
@@ -413,13 +563,19 @@ async function processDeepResearchJob(
...
+    writeStateSerialized = async (options?: ConversationStateWriteOptions) => {
+      const p = stateWriteChain.then(async () => {
+        await prepareConversationStateForWrite(options);
+        return updateConversationState(
+          activeConversationState.id!,
+          activeConversationState.values,
+        );
+      },
       );
       stateWriteChain = p.catch(() => {}); // prevent unhandled rejection from blocking chain
```
maschlr: **[P1]** Same `stateWriteChain = p.catch(() => {})` silent error swallowing as in `start.ts`. Add logging here too.

maschlr (commented):
Review with 4 specialized agents (code review, silent failure hunting, type design, code simplification). See inline comments for findings.

src/services/queue/workers/deep-research.worker.ts
raghav-rama: This is a known bug / code quality issue and will take a bigger refactor to fix properly.

Not just this, but all *.worker.ts files have duplicate code. The non-worker files (in-process mode) which were initially built were not really scalable hence we added support for BullMQ for scaling horizontally on demand afterwards.

On prod, worker mode is used and locally in-proces mode is used

raghav-rama:
I've tried to addrss all comments @maschlr 

Think we should tackle the duplicate code issue in a dedicated PR

src/services/queue/workers/deep-research.worker.ts
maschlr: @claude analyze the conversation and the problem in the codebase. Create a concise issue in github that contains all the necessary context for the proposed refactor.

raghav-rama: #157 @maschlr we have it

maschlr:
> I've tried to addrss all comments @maschlr
> 
> Think we should tackle the duplicate code issue in a dedicated PR

~~I agree. Could you create some artifact (e.g. github issue, linear issue) with the necessary context so we can line in the refactor when we have some capacity for it?~~

Also, could you resolve and/or leave a short comment on my review comments?

src/routes/deep-research/start.ts
raghav-rama: Will be fixed in future PR See #157

src/services/queue/workers/deep-research.worker.ts
raghav-rama: Will be fixed in future PR See #157

raghav-rama:
> > I've tried to addrss all comments @maschlr
> > Think we should tackle the duplicate code issue in a dedicated PR
> 
> ~I agree. Could you create some artifact (e.g. github issue, linear issue) with the necessary context so we can line in the refactor when we have some capacity for it?~
> 
> Also, could you resolve and/or leave a short comment on my review comments?

All done

maschlr (approved):
lfg!

PR #155, opened 2026-04-07
[DEV] new chat-mode, use gpt 5.4, general fixes

raghav-rama:
### Changes
- New fast chat mode by @Gautam2086 which is faster compared to previous pipeline for simple questions and for complex queries uses the literature tool. This is decided dynamically by the LLM and is also agentic so if one literature search doesn't pull enough context another may be used. So far it has only literature tool but more tools will be added to its tool arsenal starting with code execution.
- Thinking traces of the current DR on the loading screen–Small UI/UX enhancement on the UI where the reasoning traces from the current Deep Research happening are surfaced on the main loading screen
- LLMs switched to gpt-5.4 on infra deployment
- General bug fixes


raghav-rama:
No need to open new PR to main, we can work with this one just fine

maschlr:
If we would make the PR message a little bit more descriptive, this could serve directly as input for marketing.

raghav-rama:
> If we would make the PR message a little bit more descriptive, this could serve directly as input for marketing.

Done. 

Only the first two points are relevant for them

maschlr (approved):
lgtm 🚀

PR #156, opened 2026-04-09
Add basic code quality infrastructure

maschlr:
## Summary

Broad code-quality pass on the BioAgents server and client. Four main areas:

### Tooling & CI
- **Biome** replaces Prettier for lint, format, and import sorting (single tool, faster, stricter)
- **Husky pre-commit hook** runs Biome on staged files only — incremental, no repo-wide reformatting
- **GitHub Actions CI** with parallel `lint-format` (PR-changed files) and `test` (typecheck + `bun:test`) jobs
- `.editorconfig` for cross-editor consistency
- `CODING_GUIDELINES.md` documenting team principles
- `AGENTS.md` as the canonical agent-facing instructions; `CLAUDE.md` symlinked for backwards compatibility

### TypeScript strict mode
- Server and client both compile clean under strict mode; tsconfigs scoped per workspace
- Eliminated every `: any` and `as any` across `src/` and `client/`
- Removed stale `@ts-ignore` directives in LLM adapters
- Typed route handlers via a shared `ElysiaRouteContext` (`src/types/elysia.ts`)
- Structurally typed WebSocket handler
- Typed DB operations, LLM provider-name validation, and fallback handling
- Fixed all resulting regressions from the strict-mode refactor

### x402 / b402 payment protocol removed
- Deleted middleware, routes, services, scripts, migrations, DB operations, and the `X402_V2_MIGRATION.md` doc
- Removed client-side wallet hooks (`useEmbeddedWallet`, `useX402Payment`), `CDPProvider`, `PaymentConfirmationModal`, `EmbeddedWalletAuth`
- New migration `20260414000000_drop_x402_tables.sql` drops the leftover tables
- Related env vars removed from `.env.example`; auth docs and sidebar trimmed accordingly

### Error handling & silent-failure fixes
- Surfaced silent failures in `openrouter`, `fetchWithRetry`, and the status reader instead of swallowing errors
- Extracted shared body-parsing helpers into `src/utils/bodyParsing.ts` and dropped an unused type surface
- Minor resilience fixes flagged by the strict-mode refactor

### Tests
New `bun:test` coverage for core utilities and critical paths:
- `src/utils/__tests__/`: `uuid`, `cache`, `state`, `discovery`, `planningJsonExtractor`
- `src/services/paper/utils/__tests__/`: `doi`, `textUtils`
- `src/middleware/__tests__/authResolver.test.ts`
- `src/llm/__tests__/retry.test.ts`
[...]

PR #166, opened 2026-04-19
Run post-review regression tests in CI via local Supabase stack

maschlr:
## Summary

- Extends CI with a new `integration` job that spins up a Supabase local stack and Redis service container, then runs describe blocks tagged `[integration]`.
- Closes the coverage gap for three regression fixes from the parent PR (#156) that depend on real external systems: the PDF extraction error path (fix 1.2), the Supabase 23505 race in \`ensureUserAndConversation\` (fix 1.2 cont.), and the legacy \`rawFiles\` migration in \`cleanValues\` (fix 1.1).
- Integration tests skip cleanly when their required env is absent (via \`describeIf*\` helpers), so the default \`bun test\` stays safe locally.

## Why stacked on #156

The three new integration tests assert behaviors that the parent PR's fixes introduced. Basing here keeps the diff reviewable and lets the integration job verify #156's regression fixes against a real stack before either PR merges.

## Test plan

- [x] Verify CI's new \`integration\` job appears in PR checks and runs green
- [x] Pull the branch locally, run \`supabase start\` + \`bun test\` — confirm full suite passes (142 tests)
- [x] Without Supabase running, run \`bun test\` — confirm 139 pass + 5 skip, no failures
- [ ] Optional: revert the rawFiles strip in \`src/db/cleanValues.ts\` on a scratch branch; run integration job — expect test #3 to fail (proves the coverage)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,8 @@
 name: CI
 on:
+  # Run on every PR regardless of base branch so stacked PRs (e.g. targeting
+  # another feature branch) still get verified before merge.
   pull_request:
-    branches: [main, dev]
   push:
     branches: [main, dev]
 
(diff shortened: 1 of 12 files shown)


jiggyjo11 (approved):
LG !

# PR #166 — Summary for review

## What this PR does

Adds a second CI job that boots a **real Supabase + Redis stack** in GitHub Actions and runs integration tests against it. Previously CI only ran unit tests, so bugs that only surface when code actually talks to a database (race conditions, JSONB size issues, error-spread bugs) couldn't be caught pre-merge.

## Changes at a glance

### CI (`.github/workflows/ci.yml`)
- New `integration` job: Redis service container + `supabase/setup-cli` → `supabase start` (applies migrations automatically) → run tests tagged `[integration]` → `supabase stop`.
- Env vars (`SUPABASE_URL`, `SUPABASE_SERVICE_KEY`, `SUPABASE_ANON_KEY`) parsed out of `supabase status -o env` with a fail-fast check — if parsing drifts, the job errors instead of silently skipping all integration suites and going green.
- Replaced the old `docker-build` smoke job.
- CI now runs on PRs to any branch (not just `main`/`dev`) so stacked PRs get verified.

### Test infrastructure (`src/utils/__testHelpers__/integrationEnv.ts`)
- New `describeIfSupabase` / `describeIfPdf` helpers — resolve to `describe` when env is present, `describe.skip` when not.
- Guards against meaningless values like `"undefined"` / `""` / `"null"` being treated as truthy.
- Result: `bun test` locally is always safe; integration suites just skip without infra.

### New integration tests
- **`cleanValues.integration.test.ts`** — asserts `updateState` strips `buffer` and `parsedText` from `rawFiles` before persistence, so JSONB rows don't balloon as files accumulate. Seeds the row with heavy fields intact first to avoid vacuous passes.
- **`setup.integration.test.ts`** — 10 concurrent `ensureUserAndConversation` calls against the same `conversationId`; asserts exactly one winner, losers all get the standard "Access denied" string, no generic `create_conversation_failed` log, DB row belongs to the winner.
- **`description.integration.test.ts`** — feeds a corrupted PDF fixture, asserts the catch path returns a concrete error message (not the literal string `"unknown"`) and that `logger.error` fires once with a non-empty `err.message`. Guards the `{...error}` spread regression where `Error` props are non-enumerable.

### New unit test (`setup.race.test.ts`)
The Postgres `23505` unique-violation arm isn't reliably reachable from concurrency alone — HTTP round-trips serialize `createUser` so the winner usually commits before other callers reach `getConversation`. This test mocks the operations module to throw `23505` deterministically and asserts the warn log + "Access denied" return. Pairs with the integration test above (integration proves outcome, unit pins the branch).

### Production code
- `src/services/chat/setup.ts` — one new `logger.warn(..., "conversation_create_race_23505")` line so the race path is observable.
- `src/services/files/description.ts` — `extractPDFText` changed from private to `export`ed so integration tests can import it. No behavior change.

### Local dev
- `supabase/config.toml` + `supabase/.gitignore` so `supabase start` works locally.
- `AGENTS.md` updated with the local integration-test workflow.

## Open reviewer comment

`greptile-apps[bot]` flagged that `supabase/setup-cli@v1` uses `version: latest`. Since `supabase status -o env` output format and `supabase start` flag semantics have shifted across CLI releases, an unpinned CLI is a latent stability risk. Pinning to the exact version this was tested against is the safer call.

## Diff size

`+810 / -19` across 12 files. Most of the volume is the auto-generated `supabase/config.toml` (413 lines) and the new test files; the production-code delta is ~9 lines.

PR #168, opened 2026-04-21
Added SSE streaming to chat agent

Gautam2086:
## Summary
Adds Server-Sent Events streaming to `POST /api/chat`. When the client
sends `Accept: text/event-stream`, the agent loop runs in-process and
streams tokens directly over the response. Non-SSE callers fall through
to the existing queue and JSON paths unchanged.

Supersedes #165 (that version routed streaming through Redis pub/sub in
queue mode, heavier than the 5-30s chat use case needs).

## Event contract
```
init          { messageId, conversationId }
stream_start  { turnIndex }
delta         { text, turnIndex }
stream_end    { isFinal, turnIndex, reason }   // paused | complete | truncated
done          { messageId }
error         { reason, message }              // reason e.g. truncated | empty_reply | agent_error
```

`turnIndex` is on every delta so the consumer can track turn boundaries
regardless of event ordering.

## Orphan recovery
A process death mid-stream would otherwise leave a `messages` row stuck
without a reply forever, breaking history loads and the UI loading
state. This PR adds chat-mode orphan recovery while preserving
deep-research's existing BullMQ retry model. The two boundary touches
in deep-research code (enqueue-time discriminator write, branch
normalization) exist to keep that scoping safe rather than to extend
the state machine into deep research.

- **Schema:** new `status` column on `messages` with values
  `PENDING / COMPLETE / FAILED`. Migration is idempotent; backfill runs
  only on first apply.
- **Chat writers:** SSE and non-SSE catch blocks call `markMessageFailed`
  on terminal errors (guarded with `.neq("status", "COMPLETE")` so a
  post-save throw can't downgrade a saved row).
- **Reader split:** new `getCompletedMessagesByConversation` is used
  only by the chat agent runner. The general `getMessagesByConversation`
  retains pre-PR behavior (returns all rows) for deep-research callers.
- **Sweeper:** periodic background job flips PENDING rows older than 20
  min to FAILED. Scoped to chat-mode rows by reading the related
  `state.values.isDeepResearch` flag — deep-research jobs (Steering
  ~20min, Smart ~60min, Autonomous up to ~8h) are excluded so legitimate
  runs aren't clobbered. Runs as a BullMQ repeatable job in queue mode
  and `setInterval` fallback in API server when `USE_JOB_QUEUE=false`.
- **Discriminator timing:** `isDeepResearch` is written to the state row
  at enqueue time in `start.ts` (strict — fails the request if the write
  fails) so queued deep-research jobs are excluded by the sweeper before
  their worker ever runs.

## Branch normalization
Branch copies now route every source row through a
`normalizeBranchedMessage` helper that rewrites PENDING source rows to
FAILED in the branched conversation. Rationale: the copy nullifies
`state_id`, so a PENDING row in the branch can never be picked up by a
worker — marking it FAILED keeps the branch internally consistent and
preserves the user's question. Unit tests cover all three status
transitions.

## Worth knowing
- `done` only emits after the reply is durably saved and post-save
  metadata updates complete, so a client seeing it can trust the
  message is fully written.
- Truncation is caught before the DB write. Partial `max_tokens` replies
  never persist as success.
- Client disconnect clears the heartbeat but lets the agent finish;
  refreshing mid-stream surfaces the full answer on reload.
- 15s SSE comment heartbeats cover long silent stretches (tool calls,
  file processing) so proxies don't kill idle connections.
- Sweeper's `state.values.isDeepResearch` lookup is a transitional
  discriminator. Long-term cleanup is a dedicated `kind` column on
  `messages` written atomically with the row insert; tracked as a
  follow-up.

## Migration
SQL file: `supabase/migrations/20260428000000_add_message_status.sql`.
Apply manually via Supabase SQL Editor. Already applied to staging.

## Validation
- TypeScript check clean (`bun typecheck`)
- 145 unit tests pass (`bun test`) — includes new
  `normalizeBranchedMessage` coverage
- Server boots cleanly with the migration applied

## Test plan
- [x] Curl smoke tests (simple query + tool-calling query)
- [x] Browser end-to-end (first message and follow-ups both stream)
- [x] Orphan recovery: kill backend mid-stream, verify row stays
      PENDING, sweeper flips to FAILED at threshold, chat-agent history
      loader excludes FAILED rows
- [x] Sweeper deep-research safety: queued deep-research job past
      threshold is NOT swept (isDeepResearch flag set at enqueue time)

Frontend PR: https://github.com/bio-xyz/bioagent-platform-ui/pull/112

### DIFF
--- a/src/routes/chat.ts
+++ b/src/routes/chat.ts
@@ -363,30 +363,19 @@ export async function chatHandler(ctx: ElysiaRouteContext) {
     // Triggered when client sends Accept: text/event-stream header.
     // Non-SSE requests fall through to the existing queue/JSON paths below.
     // =========================================================================
-    const acceptsSSE = request.headers
-      .get("accept")
-      ?.includes("text/event-stream");
+    const acceptsSSE = request.headers.get("accept")?.includes("text/event-stream");
 
     if (acceptsSSE) {
-      logger.info(
-        { messageId: createdMessage.id, conversationId },
-        "chat_using_sse_mode",
-      );
+      logger.info({ conversationId, messageId: createdMessage.id }, "chat_using_sse_mode");
 
       // Import SSE dependencies at function scope (dynamic for TDZ safety)
       const { runChatAgent } = await import("../chat-agent/runner");
       const { fileUploadAgent } = await import("../agents/fileUpload");
-      const { getPendingFileIds, getFileStatus } = await import(
-        "../services/files/status"
+      const { getPendingFileIds, getFileStatus } = await import("../services/files/status");
+      const { getFileProcessQueue } = await import("../services/queue/queues");
+      const { getConversationState, updateMessage, updateConversationState } = await import(
+        "../db/operations"
       );
-      const { getFileProcessQueue } = await import(
-        "../services/queue/queues"
-      );
-      const {
-        getConversationState,
-        updateMessage,
-        updateConversationState,
-      } = await import("../db/operations");
 
       const encoder = new TextEncoder();
       const sseStartTime = Date.now();
@@ -395,14 +384,22 @@ export async function chatHandler(ctx: ElysiaRouteContext) {
       let heartbeatTimer: ReturnType<typeof setInterval> | null = null;
 
       const stream = new ReadableStream({
+        cancel() {
+          // Client disconnected. Agent loop keeps running via the awaited
+          // runChatAgent call - DB save still happens. send() becomes no-op.
+          // Clear the heartbeat timer to prevent leak.
+          if (heartbeatTimer) {
+            clearInterval(heartbeatTimer);
+            heartbeatTimer = null;
+          }
+          logger.info({ messageId: createdMessage.id }, "chat_sse_client_disconnected");
+        },
         async start(controller) {
           // Helper: safe enqueue (swallows errors if client disconnected)
           const send = (event: string, data: unknown) => {
             try {
               controller.enqueue(
-                encoder.encode(
-                  `event: ${event}\ndata: ${JSON.stringify(data)}\n\n`,
-                ),
+                encoder.encode(`event: ${event}\ndata: ${JSON.stringify(data)}\n\n`)
               );
             } catch {
               // Controller closed (client disconnected). Keep running so DB save happens.
@@ -416,9 +413,7 @@ export async function chatHandler(ctx: ElysiaRouteContext) {
           // (up to 2 min) can easily exceed proxy idle limits.
           const sendHeartbeat = () => {
             try {
-              controller.enqueue(
-                encoder.encode(`: heartbeat ${Date.now()}\n\n`),
-              );
+              controller.enqueue(encoder.encode(`: heartbeat ${Date.now()}\n\n`));
             } catch {
               // Controller closed
             }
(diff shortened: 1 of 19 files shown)


Gautam2086:
Thanks for the review. Addressing:

**Raw err.message exposure** — fixed in edebf37. Generic client message, full error still logs server-side via logger.error for debugging.

**No overall timeout on runChatAgent** — real issue but deserves its own focused PR (needs AGENT_LOOP_MAX_MS env var, decision on whether to abort tool loop vs whole agent loop, etc.). Tracked for follow-up.

**done.messageId unused on client** — skipping. Both IDs are always identical by construction and the handler works correctly; no user impact.

client/src/hooks/useSSEChat.ts
raghav-rama: Don't need this file

raghav-rama, line 1: Don;t need this file

client/src/pages/ChatPage.tsx
raghav-rama, line 1: Don't need this file

src/routes/chat.ts
```diff
@@ -385,6 +385,348 @@ export async function chatHandler(ctx: any) {
       "message_record_created",
     );
 
+    // =========================================================================
+    // SSE STREAMING PATH (in-process, bypasses queue mode)
```
raghav-rama, line 435: Why are we bypassing queue mode?

In-process is meant for local developement, queue mode is for prod

Also any messages I send to the agent don't show up in the bullmq dashboard

client/src/hooks/useSSEChat.ts
Gautam2086: Removed in c3e922f.

client/src/pages/ChatPage.tsx
Gautam2086: Reverted the SSE edits in c3e922f.

Gautam2086:

src/routes/chat.ts
Gautam2086: Yeah, based on what we discussed. Streaming has to bypass Bull because routing tokens through a queue would need a Redis pub/sub bus for every chunk, which is why the Vercel AI SDK and the official OpenAI and Anthropic SDKs stream directly from the request handler instead of through a queue. Deep research stays on Bull.

The visibility piece is covered by the new status column on messages along with the periodic sweeper.

src/services/queue/workers/message-sweeper.worker.ts
```diff
@@ -0,0 +1,163 @@
...
+      const state = row.state as { values?: { isDeepResearch?: boolean } } | null;
+      return state?.values?.isDeepResearch !== true;
+    })
+    .map((row) => row.id as string);
+
+  let flippedCount = 0;
+  if (chatOrphanIds.length > 0) {
+    const { data: flipped, error: updateError } = await supabase
+      .from("messages")
+      .update({ status: "FAILED" })
```
raghav-rama, line 145: If a message is in PENDING for more than 20 minutes it is marked as FAILED wrongly. 

Steps to reproduce: 
Start only the API not worker. The message will be queued. Let it sit there for 20 minutes, it will be marked failed. Now start the worker, and message is processed and marked COMPLETED.

raghav-rama (commented):
Overall a good refactor 🙌

Gautam2086:

src/services/queue/workers/message-sweeper.worker.ts
Gautam2086: Yeah, real race condition. Fixed in ca63451.

raghav-rama (approved):
Lg

PR #170, opened 2026-04-27
feat: add alphafold db support

raghav-rama:
## Summary

Adds source-selection and AlphaFold structure metadata support across orchestrator chat and deep-research flows, and forwards BioLiterature streaming tool progress through the chat SSE path.

## Key changes

- Threads `sourceSelectionId` through chat messages, state values, planning tasks, queue payloads, and Supabase schema/migrations.
- Maps selected official sources into BioLiterature `sources` overrides.
- Adds AlphaFold-specific sequence routing for explicit protein-sequence queries.
- Normalizes BioLiterature stream events into chat SSE `tool_call`, `tool_delta`, and `tool_result` events.
- Persists and returns AlphaFold protein structure metadata in normal chat, queued chat, and deep-research literature tasks.
- Adds focused tests for source routing, protein-structure extraction/persistence, BioLiterature payloads, SSE ordering, stream proxying, and chat-agent tool events.

https://github.com/user-attachments/assets/2ab0a712-627e-446f-a7b6-5dd33fae5f97


maschlr (commented):
## Review summary

Reviewed `feat/alpha-fold` (BioAgents) at `bb9ab8c` against the PR base. Inline comments cover the highest-impact findings on this PR's actual diff. A few additional notes here — some are about code already on `dev` (sibling PRs), some about code in this PR but on lines outside the changed hunks.

### Already on `dev` (sibling PRs) — flagging for whoever owns those changes

1. **`supabase/migrations/20260428000000_add_message_status.sql` — backfill clobbers in-flight rows on first deploy.** The `IF NOT EXISTS constraint` branch is the *first-run* branch — exactly the case that matters in production. New chat rows are inserted with `content: ""` (`src/services/chat/tools.ts:38`), so any chat in flight at deploy time satisfies `status = 'PENDING' AND (content IS NULL OR content = '')` and is flipped to `FAILED`. The worker's later `markMessageComplete` then no-ops on the now-FAILED row (`.eq("status","PENDING")` guard) and the user silently loses their reply. Suggest scoping the backfill to `created_at < now() - interval '30 minutes'`.

2. **`src/services/queue/workers/message-sweeper.worker.ts:100-105` — sweeper falls open silently on Redis failure.** `markerByIndex.map(() => true)` treats every PENDING row as alive forever on `pipeline.exec()` failure; only a `warn` log is emitted. A Redis hiccup silently stops orphan recovery while the system looks healthy. Promote to `error`, emit a per-run status log unconditionally with `{candidatesCount, flippedCount, redisOk}`, and check `results.length === chatOrphanIds.length` before assigning (ioredis can return partial pipeline results on transient errors).

3. **No test for `markMessageComplete`** — the load-bearing race guard from commit `ca63451` ("Guarded COMPLETE writes against post-FAILED races") is fully untested. A regression test (mock Supabase, simulate `.update().eq("status","PENDING")` returning 0 rows affected, assert no-op or `UnrecoverableError`) is the single most valuable test to add. **`message-sweeper.worker.ts` has zero tests** (241 lines).

### In this PR but on lines outside the changed hunks

4. **`src/routes/chat.ts` — `replyPersisted` shadowing.** The inner `let replyPersisted = false` declared inside `start(controller)` shadows the outer one used by the post-stream catch. Today the `.neq("status","COMPLETE")` precondition saves us, but the variable's intent is broken — anyone adding a new code path after `return new Response(stream, ...)` reintroduces a real downgrade. Remove the inner `let` and assign to the outer one.

5. **`src/routes/deep-research/branch.ts:69` — branch normalization only handles `PENDING`.** The contract per the JSDoc is "branches are snapshots, not live continuations," but the snapshot transform only normalizes PENDING. Any future status (`RUNNING`, `STREAMING`, `RETRYING`) is copied verbatim. Use an explicit terminal-status allow-list (`COMPLETE`, `FAILED`) and map everything else to `FAILED`.

6. **`src/types/core.ts:47` — `PlanTask.sources?: string[]` should be `SourceSelectionId[]`.** The Zod-backed `SourceSelectionId` union exists and is already imported into `core.ts:2`. Three fields still use loose `string[]`: `PlanTask.sources` (line 47), `ConversationStateValues.clarificationContext.initialTasks[*].sources` (line 165), and `ClarificationPlanTask.sources` (`types/clarification.ts:51`). **This is the single highest-leverage type fix in the PR** — narrowing turns typos into compile errors.

### Type-design notes

- `ProteinStructure` (`core.ts:66-82`) doesn't encode the "at least one of `bcifUrl|cifUrl|pdbUrl`" invariant — purely runtime. A discriminated union (or branded `SanitizedProteinStructure`) would prevent illegal construction.
- `messageState.ts` is misleadingly named — no states, no transitions, no narrowing. It's `Partial<StateValues>` merge plumbing. Either rename to `buildMessageStateValues` or actually model `MessageStatus = "PENDING" | "COMPLETE" | "FAILED"` as a discriminator.

### Test-coverage gaps in this PR

- `findLabeledSequence` regex has no test rejecting bare `sequence: ACGT...` (DNA misclassified as protein — see inline comment).
- `extractProteinStructuresFromBioLiteratureResponse` (`proteinStructures.ts:116-144`) only happy-path-covered transitively via `agent-stream-events.test.ts`.

Core design is sound — the `markMessageComplete` race-guard cleanly closes the COMPLETE-vs-FAILED race, AlphaFold structure normalization is well-bounded, and `fileWait.ts` test coverage is excellent.

maschlr:
## Inline-style findings (anchored to file:line)

GitHub's API was rejecting inline review comments on this PR with `pull_request_review_thread.base: An internal error occurred` (REST + GraphQL both blocked — likely a server-side bug related to the merge commit in the branch). Posting the inline content here as a follow-up so it stays anchored to specific lines.

---

### 1. `src/services/literature/agent-stream-events.ts:32-39` — Critical: malformed `data:` frame nukes the whole literature stream

`parseEventData` does `JSON.parse` with no try/catch. It's called from inside `eventsource-parser`'s `onEvent` (around line 226). Any synchronous throw from `onEvent` propagates out of `parser.feed()` in the read loop, exits the `while(true)` in `consumeLiteratureAgentStream`, and the outer caller catches it. Result: every delta already accumulated in `deltaParts` and every `proteinStructure` already mined are discarded. No `streamError` envelope is emitted to the SSE client either.

Wrap the body of `onEvent` in `try/catch`, set `streamError = err.message`, log `{ err, raw: event.data }`. A test feeding three frames where the middle one is `data: {not json}` and asserting the surrounding events are still consumed would prevent regression.

---

### 2. `src/services/literature/agent-stream-events.ts:112-115` — Type design: stringly-typed event union

`eventName: string | undefined` plus a switch on string literals (line 115) erases the per-event invariant: `tool_call` requires `toolName` + `status`, `message_delta` requires `delta`, `tool_result` requires a tool ID. Today every handler re-narrows fields via `asString`/`asObject`/`asRecord`, ~30 lines of defensive code that would disappear with a discriminated union:

```ts
type LiteratureStreamEvent =
  | { event: "tool_call"; data: { toolName: string; status: ToolStatus } }
  | { event: "message_delta"; data: { delta: string } }
  | { event: "tool_result"; data: { toolCallId: string } }
  | { event: "final"; data: { /* … */ } }
  | { event: "error"; data: { message: string } };
```

Parse once at the SSE boundary (next to `JSON.parse` in `parseEventData`), narrow into the union, then dispatch. Bonus: makes the malformed-frame fix above easier to write, and makes a future contract change a compiler error rather than a silent drop.

---

### 3. `src/services/literature/agent-stream-events.ts:258-264` — Important: empty-stream fallback hides upstream failures

If the upstream stream closes with deltas but no `final` event (mid-stream upstream crash, proxy truncation, etc.), `deltaParts.join("").trim()` is silently used as the BioLit answer with no `isError` flag and no log. The LLM will then format a partial / aborted response as if it were the canonical answer.

Distinguish "closed without final AND no streamError" from "closed with final":

```ts
if (!streamError && !sawFinal) {
  return { content: "BioLiterature stream ended unexpectedly", isError: true };
}
```

Track `sawFinal` from the `"final"` switch case in `handleEvent`.

---

### 4. `src/routes/chat.ts:731-735` — Critical: `notifyMessageUpdated` and `notifyJobCompleted` not called on the SSE success path

The queue worker calls both after `markMessageComplete` (`chat.worker.ts:405`, `:569`) so other tabs / WebSocket subscribers re-fetch. The new in-process SSE branch only emits SSE events to the originating client. Result: any other tab subscribed to this conversation channel never learns the message changed — silent state desync for multi-tab users.

Mirror the worker around the terminal-success block (after `markMessageComplete` succeeds, before `safeClose`):

```ts
try {
  await notifyMessageUpdated(conversationId, createdMessage.id);
  await notifyJobCompleted(/* … */);
} catch (err) {
  logger.warn({ err, messageId: createdMessage.id }, "chat_sse_notify_failed");
}
```

Wrap in try/catch — notify failures shouldn't fail the request.

---

### 5. `src/utils/sourceSelectionRouting.ts:104` — Important: regex matches DNA sequences as protein sequences

The trailing `|sequence` alternative makes the qualifier optional:

```ts
/\b(?:(?:protein|amino acid|aa|peptide)\s+sequence|sequence)\b\s*[:=]/gi
```

Combined with `AMINO_ACID_ALPHABET` (which contains A, C, G, T — all valid amino acids), a 20-char DNA string after a bare `"sequence:"` triggers `isExplicitProteinSequence` and the planner silently rewrites the user's question to a DNA blob.

```
What is the genome sequence: ACGTACGTACGTACGTACGT...
```

The test at `__tests__/sourceSelectionRouting.test.ts:90` only proves short identifiers don't match. Either drop the bare `|sequence` alternative (require an explicit `protein|amino acid|aa|peptide` qualifier), or filter known DNA-only alphabets before routing. Add a regression test:

```ts
expect(extractExplicitProteinSequence("sequence: " + "ACGT".repeat(20))).toBeUndefined();
```

---

### 6. `src/utils/proteinStructures.ts:33-61` — Important: silent drop with no log

`normalizeProteinStructure` returns `undefined` when (a) `entryId` is missing OR (b) all of `bcifUrl`/`cifUrl`/`pdbUrl` are absent. In both cases the structure is silently filtered out. There's no log, no metric, no signal anywhere.

A backend rename — say AlphaFold ships `bcif_url` instead of `bcifUrl`, or `entry_id` instead of `entryId` — would silently drop **every** structure. The user sees a chat answer that mentions the protein in text but with no 3D structure card, and the regression is invisible until manually noticed.

Minimum: count drops per call and emit at `warn` when the call drops everything:

```ts
logger.warn(
  { droppedCount, totalCount, sampleKeys: Object.keys(value).slice(0, 8) },
  "protein_structure_normalize_dropped_all"
);
```

Better: replace the hand-rolled `asString`/`asNumber` validators with a Zod schema and surface the parse error.

Related: the `ProteinStructure` type at `core.ts:66-82` makes every URL independently optional — encode "at least one of bcifUrl|cifUrl|pdbUrl" as a discriminated union (or return a branded `SanitizedProteinStructure`) so consumers can't accept unverified shapes.

raghav-rama:
Addressed the non-threaded inline-style findings in 8833c35: malformed Literature stream frames and handler failures now emit failed tool results; failed streams cancel promptly; missing final is an error; chat SSE completion sends subscriber notifications; bare DNA-only sequence labels no longer force AlphaFold; dropped AlphaFold candidates warn; source fields are narrowed to SourceSelectionId[]. Verified with bun test, bun run typecheck, and bun run style:check.

maschlr (approved):
lgtm

PR #171, opened 2026-04-28
returning message id for ui

jiggyjo11:
adding message id to message object when returning chat responses


maschlr (approved):
lvgtm, love the simple fix 🎯

PR #173, opened 2026-05-11
Add Kubernetes worker deployment (Kustomize + GHCR + KEDA)

maschlr:
## Summary

- Splits the single BullMQ worker process into 4 per-queue Kubernetes Deployments so each queue sizes its grace period to its own max job duration (deep-research 8h, chat 10min, heavy 1h, scheduled 2min) instead of forcing every queue to inherit the worst case.
- `src/worker.ts` adds `ENABLE_*_WORKER` env gating (defaults `true`, preserves docker-compose behaviour) and a `Bun.serve` health endpoint on `:9000` for K8s liveness/readiness probes.
- Kustomize layout under `k8s/` with `base` + `kind` / `eks-staging` (dev branch) / `eks-prod` (main branch) overlays; KEDA Redis-list scaler per queue.
- GitHub Actions: `build-image.yml` pushes to GHCR; `deploy-workers.yml` fires on build success, OIDC → per-env AWS role, pins the SHA via `kustomize edit set image`, applies.
- Design doc for review: `documentation/docs/K8S_WORKER_MIGRATION.md`.
- Existing Swarm path (`docker-compose.swarm.yml`, `deploy-worker.yml`) is untouched — both systems run in parallel during cutover.

## Test plan

- [ ] `kubectl kustomize k8s/overlays/{kind,eks-staging,eks-prod}` builds cleanly
- [ ] Local kind run: `docker build && kind load && kubectl apply -k k8s/overlays/kind` — all 4 Deployments + in-cluster Redis healthy
- [ ] For each Deployment, `kubectl exec ... env | grep ENABLE_` shows only its own queue enabled
- [ ] `curl :9000/health/ready` returns 200 with the expected `{workers, redisStatus}` payload
- [ ] KEDA: install per `k8s/README.md`; sustained `LPUSH bull:chat:wait` flood scales `worker-chat` to its overlay max
- [ ] `kubectl delete pod` mid-run: SIGTERM → clean shutdown → new pod brought up before old gone
- [ ] Before staging cutover: enqueue a job in staging Redis, confirm no prod worker consumes (and symmetric reverse)
- [ ] EKS prerequisites land (depends on Terraform workstream): cluster, namespaces, MNG, IRSA + OIDC trust scoped to refs/heads/main for prod role
- [ ] GitHub Actions vars/secrets set: `AWS_REGION`, `EKS_CLUSTER_NAME`, `AWS_ROLE_{STAGING,PROD}`, `CONFIG_ENV_{STAGING,PROD}`, `SECRET_ENV_{STAGING,PROD}`, `GHCR_PULL_PAT`


documentation/docs/K8S_WORKER_MIGRATION.md
```diff
@@ -0,0 +1,230 @@
...
+| Branch → env | `dev` → staging, `main` → prod | standard GitHub-flow promotion |
+| Image tagging | `:sha-<short>` immutable + branch tag | immutable for GitOps; branch tag for convenience |
+| Registry | GHCR (`ghcr.io/bio-xyz/bioagents`) | already on GitHub; auth via PAT-backed `imagePullSecret` |
+| IaC (EKS) | Terraform | community-mature; separate workstream from app manifests |
+| Deploy | `kubectl apply -k` from GitHub Actions | simplest path; ArgoCD when second app joins the cluster |
+| Secrets | plain k8s `Secret` from CI store | ESO + AWS Secrets Manager deferred to v1.1 |
+| Autoscaling | KEDA + Redis list scaler | only signal that reflects actual backlog |
+| Nodes | MNG + Cluster Autoscaler | predictable; Karpenter when scope grows |
+| Spot | deferred (on-demand only at launch) | de-risk cutover; cost optimisation later |
+| Logging | stdout → Fluent Bit → CloudWatch Logs | EKS default; no app change needed |
```
tjwelde: yeah, cloudwatch logs are not that great tbh. If you dont expect to have to use them regularly, it can be fine, but it will get annoying quickly.
Alternatives: grafana loki (also great together with grafana metrics), can be just added to the cluster, or eb dash0, which would be based on opentelemetry.

documentation/docs/K8S_WORKER_MIGRATION.md
```diff
@@ -0,0 +1,230 @@
...
+Two independent environments deployed from separate branches:
+
+| Env | Branch | Namespace | Redis | Supabase | Image tag |
+|---|---|---|---|---|---|
+| Staging | `dev` | `bioagents-staging` | dedicated Upstash instance | dedicated Supabase project | `:dev-sha-<short>` + `:dev` |
+| Production | `main` | `bioagents-prod` | dedicated Upstash instance | dedicated Supabase project | `:sha-<short>` + `:main` |
+
+Single EKS cluster, two namespaces. Isolation comes from separate Redis instances (BullMQ queues never cross envs) and separate Supabase projects (data never crosses). ResourceQuotas on each namespace prevent one env from starving the other. ServiceAccounts are namespace-scoped.
+
+Trade-off accepted: shared control plane upgrades, shared node-pool noisy-neighbour risk. Revisit if compliance or scale demands two clusters.
```
tjwelde, line 59: yeah, 1 control plane will be enough for a while

k8s/base/worker-scheduled/deployment.yaml
```diff
@@ -0,0 +1,103 @@
...
+                name: worker-config
+            - secretRef:
+                name: worker-secrets
+          resources:
+            requests:
+              cpu: "50m"
+              memory: "128Mi"
+            limits:
+              cpu: "200m"
+              memory: "256Mi"
```
tjwelde: careful with memory limits. if reached, this will do a hard shutdown of the pod. 
That is in contrast to cpu limits, where the cpu just gets throttled.

k8s/base/worker-scheduled/deployment.yaml
```diff
@@ -0,0 +1,103 @@
...
+            periodSeconds: 5
+          readinessProbe:
+            httpGet:
+              path: /health/ready
+              port: health
+            periodSeconds: 10
+            failureThreshold: 3
+          livenessProbe:
+            httpGet:
+              path: /health/live
```
tjwelde, line 94: will the worker be able to answer, while processing? if it cant, k8s will restart the pod

.github/workflows/deploy-workers.yml
```diff
@@ -0,0 +1,120 @@
...
+          set -eu
+          AUTH=$(printf "%s:%s" "$GHCR_USER" "$GHCR_PAT" | base64 -w 0)
+          cat > k8s/overlays/${{ steps.env.outputs.overlay }}/pull-secret.dockerconfigjson <<EOF
+          {"auths":{"ghcr.io":{"auth":"$AUTH"}}}
+          EOF
+
+      - name: Install kustomize
+        run: |
+          set -eu
+          curl -fsSL "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh" | bash
```
tjwelde: kustomize should be part of the github action `ubuntu-latest` image. or do we need a different version?

documentation/docs/K8S_WORKER_MIGRATION.md
maschlr: Switched to in-cluster Loki + Grafana Alloy DaemonSet (`grafana/loki` simple-scalable, S3 chunks via IRSA). Installed by Terraform — `infra/terraform/modules/observability/`. dash0 was tempting but Loki keeps things self-hosted and pairs naturally with the Prometheus/Grafana stack we'll add for metrics later. Commit 4c97c08 (decision) + 8927c0e (TF wiring).

documentation/docs/K8S_WORKER_MIGRATION.md
maschlr: 👍 Locked in: single EKS cluster, `bioagents-staging` + `bioagents-prod` namespaces. Terraform composes both envs from the same modules in `infra/terraform/envs/`.

k8s/base/worker-scheduled/deployment.yaml
maschlr: Good catch — dropped memory limits on all four worker Deployments. OOMKill is a hard, signal-less termination that bypasses `worker.close()`, which would have silently broken our in-flight-job drain guarantee. CPU limits stay (they throttle, not kill). `worker-deep-research` moves from Guaranteed → Burstable QoS as a result; `safe-to-evict: false` still prevents autoscaler scale-down mid-job. Commit 4c97c08.

k8s/base/worker-scheduled/deployment.yaml
maschlr: Real concern. `Bun.serve` shares the event loop with BullMQ. Tuned the probe to `timeoutSeconds: 5, failureThreshold: 10, periodSeconds: 30` — 300s budget before a restart. Most jobs are I/O-bound (LLM HTTP, Supabase, S3) and yield to the loop; Pandoc/LaTeX in `worker-heavy` runs as a subprocess via `Bun.spawn` so it doesn't block the parent. The remaining risk is `file-process` (OCR can block briefly) — the 300s budget should cover it. BullMQ's own stall detection (lock extension) handles deadlock; the K8s liveness probe is the long-tail fallback for cases BullMQ can't catch. Documented the event-loop assumption in `documentation/docs/K8S_WORKER_MIGRATION.md` under "Liveness probe & event-loop assumption". Commit 4c97c08.

.github/workflows/deploy-workers.yml
maschlr: You're right — `ubuntu-latest` ships it. Dropped the step. Commit 4c97c08.

.github/workflows/deploy-workers.yml
```diff
@@ -0,0 +1,115 @@
...
+          kustomize build k8s/overlays/${{ steps.env.outputs.overlay }} \
+            | kubectl apply --server-side --force-conflicts -f -
+
+      - name: Wait for rollout
+        run: |
+          set -eu
+          NS=$([ "${{ steps.env.outputs.overlay }}" = "eks-prod" ] && echo bioagents-prod || echo bioagents-staging)
+          # Long timeout because deep-research drain can be up to 8h.
+          for d in worker-deep-research worker-chat worker-heavy worker-scheduled; do
+            kubectl -n "$NS" rollout status deploy/"$d" --timeout=30000s
```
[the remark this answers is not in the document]
maschlr: Switched the wait step from `kubectl rollout status --timeout=30000s` to `kubectl wait --for=condition=Available --timeout=30m`. `Available` flips true as soon as the new ReplicaSet has its pod Ready (under `maxUnavailable: 0`/`maxSurge: 100%`), so we don't block on the old pod's drain — which for `worker-deep-research` can run up to 8h and would always exceed GHA's 6h job limit. Old pods finish draining asynchronously. Also bounded the job with `timeout-minutes: 120`. Commit f0eb8f6.

.github/workflows/deploy-workers.yml
```diff
@@ -0,0 +1,115 @@
...
+
+      - name: Configure AWS credentials (OIDC)
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          # Per-env role; trust policy must scope the prod role to refs/heads/main.
+          role-to-assume: ${{ steps.env.outputs.overlay == 'eks-prod' && secrets.AWS_ROLE_PROD || secrets.AWS_ROLE_STAGING }}
+          aws-region: ${{ env.AWS_REGION }}
+
+      - name: Update kubeconfig
+        run: aws eks update-kubeconfig --name "$CLUSTER_NAME" --region "$AWS_REGION"
```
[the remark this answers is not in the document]
maschlr: Added a `Verify required secrets for overlay` step that fails fast with `::error::Missing required secrets for overlay <X>: …` when the per-env secret is empty. The `&&/||` ternaries still appear in the downstream steps but are guarded — the validation step runs first and exits non-zero before any substitution can happen, so a missing `AWS_ROLE_PROD` aborts the deploy instead of silently falling back to staging. Commit f0eb8f6.

.github/workflows/deploy-workers.yml
```diff
@@ -0,0 +1,115 @@
...
+      - name: Render pull-secret
+        env:
+          GHCR_USER: ${{ github.repository_owner }}
+          GHCR_PAT: ${{ secrets.GHCR_PULL_PAT }}
+        run: |
+          set -eu
+          AUTH=$(printf "%s:%s" "$GHCR_USER" "$GHCR_PAT" | base64 -w 0)
+          cat > k8s/overlays/${{ steps.env.outputs.overlay }}/pull-secret.dockerconfigjson <<EOF
+          {"auths":{"ghcr.io":{"auth":"$AUTH"}}}
+          EOF
```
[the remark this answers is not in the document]
maschlr: Applied. Replaced the heredoc with `umask 077` + `printf` + `chmod 400` so the GHCR PAT file is unreadable to subsequent steps. Commit f0eb8f6.

infra/terraform/modules/deployer-role/main.tf
```diff
@@ -0,0 +1,75 @@
...
+  principal_arn = aws_iam_role.deployer.arn
+
+  # AmazonEKSClusterAdminPolicy is the broadest. Tighten later by switching
+  # to AmazonEKSAdminPolicy + a namespace scope if cluster-admin is too much.
+  policy_arn = "arn:aws:eks::aws:cluster-access-policy/AmazonEKSClusterAdminPolicy"
+
+  access_scope {
+    type = "cluster"
+  }
+}
```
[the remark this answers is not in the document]
maschlr: Tightened. Switched from `AmazonEKSClusterAdminPolicy` at cluster scope to `AmazonEKSAdminPolicy` at namespace scope (new required `target_namespace` module variable; `envs/{staging,prod}` pass `bioagents-<env_name>`). Sufficient for the deploy workflow — KEDA controller, Loki, etc. are managed by Terraform, not by this role. Commit f0eb8f6.

.github/workflows/deploy-workers.yml
```diff
@@ -0,0 +1,115 @@
...
+        required: false
+
+permissions:
+  id-token: write
+  contents: read
+
+concurrency:
+  # Prevent overlapping deploys per env.
+  group: deploy-workers-${{ github.ref }}
+  cancel-in-progress: false
```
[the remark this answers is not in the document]
maschlr: Added `timeout-minutes: 120` to the deploy job. Combined with the rollout-wait change (now `kubectl wait --for=condition=Available --timeout=30m`), nothing in the job can sit longer than ~30min on the long-tail rollout — well under GHA's 6h limit. Commit f0eb8f6.

tjwelde (approved):
LGTM

PR #180, opened 2026-05-13
feat/segment-anything

raghav-rama:
## Summary
- Add normal-chat handling for `toolId=segment-anything`
- Validate uploaded image inputs, call BioLiterature SAM, and persist annotated output as a chat artifact
- Propagate artifacts through sync, SSE, queued chat, notifications, and conversation state
- Add artifact merge helpers and focused route/service tests
- Related to bio-xyz/bioagent-platform-ui PR 129

## Screenshots
image


src/routes/chat.ts
```diff
@@ -753,7 +877,7 @@ export async function chatHandler(ctx: ElysiaRouteContext) {
     }
 
     const { set } = ctx;
-    set.status = 500;
+    set.status = error instanceof SegmentAnythingToolError ? error.statusCode : 500;
```
maschlr, line 864: **Critical: upstream BioLiterature error text leaked to clients (non-SSE path).** This returns `err.message` verbatim with `SegmentAnythingToolError.statusCode`. `callBioLiteratureSegmentAnything` (`chat-tool.ts:179-184`) builds messages like ``BioLiterature Segment Anything error: ${status} - ${errorText}`` from the upstream response body — can leak stack traces, internal hostnames, or auth diagnostics.

The SSE catch in `chat-sse-transport.ts` already uses a generic message; match it here. Log the raw error, return `"Segment Anything failed"` for `statusCode >= 500` (keep the user-facing 4xx messages such as the prompt-length / image-size guards).

src/services/segment-anything/chat-tool.ts
```diff
@@ -0,0 +1,298 @@
...
+  datasets: UploadedDataset[],
+  input: SegmentAnythingToolInput
+): UploadedDataset | undefined {
+  if (input.imageFileId) {
+    return datasets.find((dataset) => dataset.id === input.imageFileId);
+  }
+  if (input.imageFilename) {
+    return datasets.find((dataset) => dataset.filename === input.imageFilename);
+  }
+  return datasets.find((dataset) => isImageUpload(dataset.filename));
```
maschlr: **Critical: silent wrong-image selection.** When both `imageFileId` and `imageFilename` are absent and the conversation has multiple uploaded images, this picks the first match with no signal. SAM runs against the wrong picture; the user only notices by reading the filename embedded in the summary.

Fix: throw `SegmentAnythingToolError("Specify which image to segment")` when both fields are absent **and** `datasets.filter(d => isImageUpload(d.filename)).length > 1`. Only allow the implicit pick when exactly one image exists.

Cross-PR: the frontend ([bio-xyz/bioagent-platform-ui#129](https://github.com/bio-xyz/bioagent-platform-ui/pull/129)) currently always sends `imageFileId` via `buildSegmentAnythingToolInput`, but defense-in-depth here protects against future tool-call paths and bad client states.

src/services/queue/workers/chat.worker.ts
```diff
@@ -43,6 +45,29 @@ async function failJobIfRowNoLongerPending(
...
+    const { updateConversationState } = await import("../../../db/operations");
+    const nextValues = withNormalChatArtifacts(
+      params.conversationState.values,
+      params.messageId,
+      params.artifacts
+    );
+    await updateConversationState(params.conversationState.id, nextValues);
+    params.conversationState.values = nextValues;
+  } catch (err) {
+    logger.warn({ error: err, jobId: params.jobId, messageId: params.messageId }, params.logKey);
```
maschlr: **Critical: artifact persistence failure is silently swallowed; client is told success.** Same pattern in `routes/chat-sse-transport.ts:91-111` and `routes/chat.ts:57-77`.

If `updateConversationState` throws, we `logger.warn` and return — but `notifyJobCompleted` (line ~245), `sendFinal({ artifacts })` (SSE), and the in-process JSON response all still ship `artifacts` to the client. On reload, `normalChatArtifactsByMessageId` is empty → the artifact URL 404s → indistinguishable from data loss with no error to explain why.

Fix: re-throw on persist failure here (and at the other two sites), or stop emitting `artifacts` in the completion payload when the write failed. For the worker, throwing here lets BullMQ mark the job failed and the user can retry via `/api/chat/retry/:jobId`.

This also masks a separate concern: three sites do read-modify-write on `normalChatArtifactsByMessageId` without the Redis lock used by `addFileWithLock` in `services/files/index.ts`. Concurrent writers lose entries because `conversationState.values` is stale by UPDATE time. Route artifact persistence through the same lock once you address the throw-vs-swallow above.

Cross-PR: frontend ([bio-xyz/bioagent-platform-ui#129](https://github.com/bio-xyz/bioagent-platform-ui/pull/129)) has no way to detect this state today.

src/services/segment-anything/chat-tool.ts
```diff
@@ -0,0 +1,298 @@
...
+
+  const fileStatus = input.imageFileId
+    ? await (deps.getFileStatus || getFileStatus)(input.imageFileId)
+    : null;
+  const contentType = fileStatus?.contentType;
+  if (!isImageUpload(dataset.filename, contentType)) {
+    throw new SegmentAnythingToolError("Segment Anything requires an image upload.");
+  }
+  const imageSize = fileStatus?.size ?? dataset.size;
+  if (typeof imageSize === "number" && imageSize > MAX_SEGMENT_ANYTHING_IMAGE_BYTES) {
```
maschlr, line 265: **Important: 50 MB size check is skipped when `fileStatus` is null.** `fileStatus` is only fetched when `input.imageFileId` is provided. With filename-fallback or the "first image" branch, `imageSize` falls back to `dataset.size` which may be `undefined` for legacy or in-message uploads — the 50 MB guard becomes a no-op, and an arbitrarily large image (up to the 2 GB upload cap) is base64-encoded into the SAM JSON POST.

Fix: when `imageSize === undefined`, either fail-closed (require a sized record) or check `rawImage.length > MAX_SEGMENT_ANYTHING_IMAGE_BYTES` post-download before encoding.

src/services/segment-anything/chat-tool.ts
```diff
@@ -0,0 +1,298 @@
...
+  request: SegmentAnythingRequest
+): Promise<SegmentAnythingResponse> {
+  const baseUrl = process.env.BIO_LIT_AGENT_API_URL?.replace(/\/$/, "");
+  const apiKey = process.env.BIO_LIT_AGENT_API_KEY || "";
+
+  if (!baseUrl || !apiKey) {
+    throw new SegmentAnythingToolError("BioLiterature API URL or API key not configured", 503);
+  }
+
+  const { response } = await fetchWithRetry(
```
maschlr: **Important: unbounded retry pins workers ~5 min per attempt × 3 BullMQ attempts.** `fetchWithRetry` defaults to 10 retries with up to ~5 min total backoff. This call doesn't override `maxRetries`, `initialDelayMs`, `maxDelayMs`, or pass an `AbortSignal`.

With `CHAT_QUEUE_CONCURRENCY=5` (default) a degraded BioLiterature endpoint starves all chat workers for ~15 minutes per stuck job.

Fix: pass `maxRetries: 2-3` and wire an `AbortController` wall-clock timeout (~60s) specifically for SAM.

src/services/chat/fileMetadata.ts
```diff
@@ -0,0 +1,59 @@
...
+  );
+}
+
+export function parseUploadedFileReferences(value: unknown): UploadedFileReference[] {
+  if (typeof value !== "string" || !value.trim()) return [];
+  try {
+    const parsed = JSON.parse(value) as unknown;
+    return Array.isArray(parsed) ? parsed.filter(isUploadedFileReference) : [];
+  } catch {
+    return [];
```
maschlr: **Important: silent JSON parse failure.** `catch {} → return []` hides malformed `fileReferences`. Chat proceeds with zero file references even though the client believed files were uploaded — user asks "summarise the PDF I just attached" → agent hallucinates with no log line.

The project CLAUDE.md explicitly forbids this pattern. At minimum `logger.warn({ err }, "parseUploadedFileReferences_invalid_json")`; ideally return `null` so the caller can 400 the request rather than silently dropping context.

src/routes/chat-notifications.ts
```diff
@@ -1,19 +1,22 @@
...
   try {
     await notifyMessageUpdated(jobId, params.conversationId, params.messageId);
-    await notifyJobCompleted(jobId, params.conversationId, params.messageId, undefined, {
-      proteinStructures: params.proteinStructures,
-    });
+    const result = {
+      ...(params.artifacts?.length ? { artifacts: params.artifacts } : {}),
+      ...(params.proteinStructures?.length ? { proteinStructures: params.proteinStructures } : {}),
+    };
+    await notifyJobCompleted(jobId, params.conversationId, params.messageId, undefined, result);
```
maschlr, line 18: **Important: single try/catch wraps two independent notifications.** If `notifyMessageUpdated` throws, `notifyJobCompleted` never fires — listening clients never receive the completion event for a reply that IS durably saved.

Fix: split into two try blocks, or push the artifacts/proteinStructures completion (`notifyJobCompleted`) **before** the messageUpdated nudge so the more important signal is delivered first.

src/services/segment-anything/chat-tool.ts
```diff
@@ -0,0 +1,298 @@
...
+  DataArtifact,
+  SegmentAnythingPoint,
+  SegmentAnythingToolInput,
+} from "../../types/core";
+import { fetchWithRetry } from "../../utils/fetchWithRetry";
+import logger from "../../utils/logger";
+import { getFileStatus } from "../files/status";
+
+const SEGMENT_ANYTHING_TOOL_ID: ChatToolId = "segment-anything";
+const IMAGE_EXTENSIONS = new Set(["bmp", "gif", "jpeg", "jpg", "png", "tif", "tiff", "webp"]);
```
maschlr: **Cross-cutting: image-extension allowlist mismatch with frontend.**

- Backend (this file): `bmp/gif/jpeg/jpg/png/tif/tiff/webp`. `imageExtensionForMimeType` (line 122) only knows `jpeg/png/webp/gif` and silently falls back to `.png` for everything else — so a TIFF input would land at S3 with `.png` extension.
- Frontend ([bio-xyz/bioagent-platform-ui#129](https://github.com/bio-xyz/bioagent-platform-ui/pull/129) `lib/segment-anything.ts:48`): `png/jpg/jpeg/webp/gif/svg` — SVG passes via content-type but TIFF/BMP are never selectable in the composer.

Decide on a single canonical set (what does the SAM model actually accept?), then align both sides. Right now: TIFF passes backend but not frontend; SVG passes frontend but extension-fallback rejects on backend.

src/services/segment-anything/__tests__/chat-tool.test.ts
```diff
@@ -0,0 +1,278 @@
+import { describe, expect, test } from "bun:test";
```
maschlr, line 1: **Test coverage — top untested risks** (full list from automated analysis):

1. SSE transport SAM branch (`chat-sse-transport.ts:271-363`): tool_call → tool_result ordering, `markReplyPersisted` flagging, the `updated:false` race path.
2. Worker `runSegmentAnythingForJob` (`chat.worker.ts:204-261`): `UnrecoverableError` on row-no-longer-pending; notify failure swallow.
3. `callBioLiteratureSegmentAnything` (`chat-tool.ts:151-187`): status remapping (5xx→502), missing-config 503.
4. `normalizePoint` (`chat-tool.ts:85-96`): boundary rejection at `x > 1` is uncovered.
5. `fetchFileByRelativePath` resolving `undefined` → expects 500.
6. Wrong-owner rejections in `requestUploadUrl` / `confirmUpload` / `deleteFile` — the cross-tenant security boundary is entirely unverified.
7. `fetchWithRetry` network-error retry + exhaustion path.

DI seams in `chat-tool.ts` and `chat-sse-transport.ts` are well-designed — these gaps are about exercising existing seams, not adding new ones.

maschlr (changes requested):
Reviewed the SAM-only commits ahead of dev. Summary: **3 critical**, **5 important**, **2 cross-cutting** with the frontend PR ([bio-xyz/bioagent-platform-ui#129](https://github.com/bio-xyz/bioagent-platform-ui/pull/129)), plus several test gaps.

Criticals are merge-blocking — the artifact-persist swallow in particular causes silent data loss: clients see an annotated image, reload, and the artifact is gone. Inline comments below pinpoint each issue and the smallest fix.

src/services/segment-anything/chat-tool.ts
```diff
@@ -0,0 +1,266 @@
...
+  const segmentClient = deps.segmentClient || callBioLiteratureSegmentAnything;
+  const segmentResponse = await segmentClient({
+    confidence: input.confidence,
+    image_base64: rawImage.toString("base64"),
+    point: input.point,
+    prompt: message,
+  });
+
+  const artifactMimeType = segmentResponse.annotated_image.mime_type || "image/png";
+  const artifactPath = `artifacts/${messageId}/segment-anything-annotated.png`;
```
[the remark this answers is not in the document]
raghav-rama: Fixed in 2f00a01. Annotated outputs now use the returned MIME type to choose the stored extension, and unsupported annotated image MIME types fail instead of being saved with a misleading extension.

src/services/segment-anything/chat-tool.ts
raghav-rama: Fixed in 2f00a01. When multiple candidate images exist and no id or filename is provided, the tool asks the user to specify the image; implicit selection is allowed only with one candidate.

src/routes/chat.ts
raghav-rama: Fixed in 2f00a01. Upstream error bodies are no longer exposed to API clients; the raw body is logged and upstream 5xx responses return Segment Anything failed.

src/services/segment-anything/chat-tool.ts
```diff
@@ -0,0 +1,266 @@
...
+  const rawImage = fileStatus?.s3Key
+    ? await storageProvider.download(fileStatus.s3Key)
+    : await storageProvider.fetchFileByRelativePath?.(
+        userId,
+        conversationState.id,
+        dataset.path || `uploads/${dataset.filename}`
+      );
+  if (!rawImage) {
+    throw new SegmentAnythingToolError("Unable to read the uploaded image", 500);
+  }
```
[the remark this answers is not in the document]
raghav-rama: Fixed in 2f00a01. The fallback path now fails explicitly when relative-path reads are unavailable instead of silently calling an optional method, and the behavior is covered by tests.

src/services/segment-anything/chat-tool.ts
```diff
@@ -0,0 +1,266 @@
...
+}
+
+function extension(filename: string): string {
+  return filename.split(".").pop()?.toLowerCase() || "";
+}
+
+function isImageUpload(filename: string, contentType?: string): boolean {
+  if (contentType) return contentType.startsWith("image/");
+  return IMAGE_EXTENSIONS.has(extension(filename));
+}
```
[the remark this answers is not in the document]
raghav-rama: Fixed in 2f00a01. SAM image detection now uses a canonical PNG/JPG/JPEG/WebP allowlist with MIME plus extension fallback and rejects SVG or unsupported raster types before the upstream call.

src/services/segment-anything/chat-tool.ts
```diff
@@ -0,0 +1,266 @@
...
+    }
+  );
+
+  if (!response.ok) {
+    const errorText = await response.text();
+    throw new SegmentAnythingToolError(
+      `BioLiterature Segment Anything error: ${response.status} - ${errorText}`,
+      response.status >= 500 ? 502 : response.status
+    );
+  }
```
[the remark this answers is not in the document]
raghav-rama: Fixed in 2f00a01. Upstream bodies are logged with status but omitted from client-facing errors, and chat returns a generic Segment Anything failed message for upstream 5xx while preserving local 4xx validation messages.

src/services/segment-anything/chat-tool.ts
raghav-rama: Fixed in 2f00a01. The tool now enforces the 50 MB limit after download whenever size metadata is missing, before base64 encoding the image.

src/services/segment-anything/chat-tool.ts
raghav-rama: Fixed in 2f00a01. Backend SAM validation now uses the same canonical PNG/JPG/JPEG/WebP set as the frontend PR and rejects SVG, TIFF, BMP, and GIF for this tool.

src/services/chat/fileMetadata.ts
raghav-rama: Fixed in 2f00a01. Malformed fileReferences now logs and causes the chat route to return 400 instead of silently dropping uploaded file context.

maschlr (approved):

PR #182, opened 2026-05-21
fix: DR failure paths now flip message row to FAILED (BIOS-94)

Gautam2086:
Closes BIOS-94.

## Summary

Deep-research failure paths were marking only the state row as `failed`; they were not calling `markMessageFailed` for the corresponding `messages` row. Because the orphan-recovery sweeper intentionally excludes deep-research rows, a crashed or failed DR run could leave a message permanently stuck in `PENDING`. Frontend consumers would then see the run as still pending indefinitely.

This PR adds `markMessageFailed(messageId)` to all four deep-research failure paths:

- `handleDeepResearchStartFailure` in `start.ts` — covers in-process runtime failures
- Queue-mode enqueue error in `start.ts` — covers failures before the worker picks up the job
- In-process start error in `start.ts` — covers early startup failures before the background run is fully underway
- Worker error handler in `deep-research.worker.ts` — covers worker-side failures on the final attempt

Each call is wrapped in its own `try/catch`, so a failure updating the `messages` table does not block the surrounding error cleanup. The existing `markMessageFailed` guard (`.neq("status", "COMPLETE")`) still prevents downgrading a row that has already completed. `normalizeBranchedMessage` remains a defensive layer for branched conversations.

## Test plan

- Trigger a deep-research failure (for example, kill the worker mid-run or force an error) and verify that the `messages` row flips to `FAILED` instead of staying `PENDING`
- Verify that a successful deep-research run is unaffected

### DIFF
--- a/src/routes/deep-research/start.ts
+++ b/src/routes/deep-research/start.ts
@@ -200,11 +200,15 @@ async function handleDeepResearchStartFailure(
     }
   }
 
-  await deps.updateState(stateRecord.id, {
-    ...stateRecord.values,
-    error: errorMessage,
-    status: "failed",
-  });
+  try {
+    await deps.updateState(stateRecord.id, {
+      ...stateRecord.values,
+      error: errorMessage,
+      status: "failed",
+    });
+  } catch (updateErr) {
+    deps.logger.error({ updateErr }, "deep_research_update_state_on_failure_failed");
+  }
 
   try {
     await markMessageFailed(rootMessageId);
--- a/src/routes/deep-research/start.ts
+++ b/src/routes/deep-research/start.ts
@@ -115,6 +115,7 @@ type DeepResearchStartFailureDeps = {
     values: ConversationState["values"],
     fallbackObjective?: string
   ) => string | undefined;
+  markMessageFailed: (messageId: string) => Promise<void>;
   markObjectiveTraceStale: (values: ConversationState["values"]) => unknown;
   updateConversationState: (id: string, values: ConversationState["values"]) => Promise<unknown>;
   notifyStateUpdated: (jobId: string, conversationId: string, stateId: string) => Promise<unknown>;
@@ -131,6 +132,7 @@ type DeepResearchStartFailureDeps = {
 
 type DeepResearchStartFailureParams = {
   activeConversationState: ConversationState | null;
+  activeMessageId?: string;
   conversationId: string;
   conversationStateId: string;
   err: unknown;
@@ -147,6 +149,7 @@ const deepResearchStartFailureDeps: DeepResearchStartFailureDeps = {
   ensureObjectiveTrace,
   getObjectiveTraceObjective,
   logger,
+  markMessageFailed,
   markObjectiveTraceStale,
   markRunFinished,
   notifyStateUpdated,
@@ -160,6 +163,7 @@ async function handleDeepResearchStartFailure(
 ): Promise<void> {
   const {
     activeConversationState,
+    activeMessageId,
     conversationId,
     conversationStateId,
     err,
@@ -211,9 +215,23 @@ async function handleDeepResearchStartFailure(
   }
 
   try {
-    await markMessageFailed(rootMessageId);
+    await deps.markMessageFailed(rootMessageId);
   } catch (msgErr) {
     deps.logger.warn({ msgErr, rootMessageId }, "deep_research_mark_message_failed_on_failure");
+  }
+
+  // If auto-continuation created a new message row before the failure, mark
+  // that row FAILED too. markMessageFailed guards .neq("status", "COMPLETE")
+  // so calling it on an already-completed root is safe.
+  if (activeMessageId && activeMessageId !== rootMessageId) {
+    try {
+      await deps.markMessageFailed(activeMessageId);
+    } catch (msgErr) {
+      deps.logger.warn(
+        { activeMessageId, msgErr },
+        "deep_research_mark_continuation_message_failed_on_failure"
+      );
+    }
   }
 
   try {
@@ -386,6 +404,7 @@ export async function deepResearchStartHandler(ctx: ElysiaRouteContext) {
   let createdMessage: CreatedMessage | null = null;
   let runMarkedStarted = false;
   let activeConversationState: ConversationState | null = null;
+  let queueJobEnqueued = false;
 
   // Log with state IDs now that we have them
   logger.info(
@@ -707,6 +726,7 @@ export async function deepResearchStartHandler(ctx: ElysiaRouteContext) {
           jobId: createdMessage.id, // Use message ID as job ID for easy lookup
         }
       );
+      queueJobEnqueued = true;
 
       activeConversationState = {
         id: conversationStateRecord.id,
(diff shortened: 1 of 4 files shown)


Gautam2086:
@greptileai review

src/routes/deep-research/start.ts
```diff
@@ -206,6 +206,13 @@ async function handleDeepResearchStartFailure(
     status: "failed",
   });
 
+  try {
+    const { markMessageFailed } = await import("../../services/chat/tools");
+    await markMessageFailed(rootMessageId);
+  } catch (msgErr) {
+    deps.logger.warn({ msgErr, rootMessageId }, "deep_research_mark_message_failed_on_failure");
+  }
```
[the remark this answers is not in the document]
Gautam2086: Fixed in commit efa930e -- markMessageFailed is now a static import alongside createMessageRecord at the top of start.ts. No dynamic import on this path.

src/services/queue/workers/deep-research.worker.ts
```diff
@@ -1380,6 +1380,13 @@ async function processDeepResearchJob(
           status: "failed",
         });
 
+        try {
+          const { markMessageFailed } = await import("../../../services/chat/tools");
+          await markMessageFailed(messageId);
+        } catch (msgErr) {
+          logger.warn({ messageId, msgErr }, "deep_research_worker_mark_message_failed_on_failure");
+        }
```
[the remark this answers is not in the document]
Gautam2086: Valid catch. Fixed in a2714a3 -- updateState is now in its own try/catch in the worker, so markMessageFailed runs independently even if the state write fails. Also applied the same fix to handleDeepResearchStartFailure in start.ts which had the same pattern.

Gautam2086:
@greptileai review

src/routes/deep-research/start.ts
```diff
@@ -200,11 +200,21 @@ async function handleDeepResearchStartFailure(
...
   }
 
-  await deps.updateState(stateRecord.id, {
-    ...stateRecord.values,
-    error: errorMessage,
-    status: "failed",
-  });
+  try {
+    await deps.updateState(stateRecord.id, {
+      ...stateRecord.values,
```
raghav-rama: This still only marks the root message failed. In the in-process auto-continuation path, `runContinuationPrepPhase` creates a new agent message and assigns it to `currentMessage`; if the next iteration then throws before its reply phase completes, the catch still calls `handleDeepResearchStartFailure` with `rootMessageId`. That means this new continuation row can remain `PENDING` while the original/root row is the one targeted here. Please pass/track the active message id for cleanup and mark that row failed as well, without downgrading already-completed rows.

src/services/queue/workers/deep-research.worker.ts
```diff
@@ -1379,33 +1379,44 @@ async function processDeepResearchJob(
...
           status: "failed",
         });
+      } catch (updateErr) {
+        logger.error({ updateErr }, "failed_to_update_state_on_error");
+      }
 
-        await clearConversationActivity({ staleTrace: true });
+      try {
+        const { markMessageFailed } = await import("../../../services/chat/tools");
+        await markMessageFailed(messageId);
```
raghav-rama, line 1403: This worker failure path has the same continuation gap. The worker creates `agentMessage` before `queue.add` for the next iteration; if `queue.add` throws, this catch runs for the current job and marks only `messageId`. That current message was already completed by `runReplyPhase`, while the newly created `nextMessageId` has no BullMQ job and is never marked `FAILED`. Track the continuation message id across this section and fail it when enqueueing the continuation fails.

src/services/queue/workers/deep-research.worker.ts
```diff
@@ -1379,33 +1379,44 @@ async function processDeepResearchJob(
...
+      } catch (msgErr) {
+        logger.warn({ messageId, msgErr }, "deep_research_worker_mark_message_failed_on_failure");
+      }
 
-        // Notify: Job failed
-        await notifyJobFailed(job.id!, conversationId, messageId, stateId);
+      await clearConversationActivity({ staleTrace: true });
```
raghav-rama: Terminal failure bookkeeping is still gated by this cleanup call. `clearConversationActivity` persists conversation state and can throw; if it does, execution skips `notifyJobFailed`, `markRunFinished`, and the credit refund block below. On a final failed attempt that leaves the run unfinished and credits un-refunded. Wrap this cleanup in its own best-effort `try/catch` before continuing to terminal bookkeeping.

src/routes/deep-research/start.ts
```diff
@@ -779,6 +789,14 @@ export async function deepResearchStartHandler(ctx: ElysiaRouteContext) {
           "deep_research_state_mark_failed_on_queue_error"
         );
       }
+      try {
+        await markMessageFailed(createdMessage.id);
```
raghav-rama: This catch covers more than enqueue failure. `queue.add` succeeds before the later activity/updateRunJobId writes; if one of those post-enqueue writes throws, the BullMQ job is already live but this marks the message `FAILED`. The worker later persists replies with `markMessageComplete`, which only updates rows still `PENDING`, so a successful live job can be prevented from completing its message row. Split the catch so only pre-enqueue/enqueue failures mark the message failed, or compensate/cancel the already-created job before flipping the row.

raghav-rama (changes requested):
Requesting changes based on the inline comments. Main issues: continuation failures can still orphan a PENDING message in both in-process and queue modes; worker final-failure cleanup can still skip markRunFinished/refund; and the queue start catch can mark a live enqueued job message FAILED after post-enqueue writes fail. Also, this PR adds no regression coverage for the claimed failure paths or the continuation/post-enqueue edge cases, so please add tests around these paths before merging.

Gautam2086:

src/routes/deep-research/start.ts
Gautam2086: Fixed in commit ef1f2ef. `handleDeepResearchStartFailure` now accepts an optional `activeMessageId` param. The in-process catch passes `currentMessage.id` (which diverges from `rootMessageId` after auto-continuation). The handler marks both IDs failed independently, each in its own try/catch, and the guard `activeMessageId !== rootMessageId` prevents double-marking when there's no continuation.

src/services/queue/workers/deep-research.worker.ts
Gautam2086: Fixed in commit ef1f2ef. The `queue.add` for the next iteration is now wrapped in its own try/catch. On enqueue failure, `markMessageFailed(nextMessageId)` is called before re-throwing, so the orphaned continuation row is immediately flipped to FAILED. The current message (`messageId`) was already completed by `runReplyPhase` so it's intentionally not touched here.

src/services/queue/workers/deep-research.worker.ts
Gautam2086: Fixed in commit ef1f2ef. `clearConversationActivity` and `notifyJobFailed` are now each wrapped in their own try/catch with warn-level logging on failure. Terminal bookkeeping (`markRunFinished`, credits refund) runs unconditionally regardless of whether those cleanup steps succeed.

src/routes/deep-research/start.ts
Gautam2086: Fixed in commit ef1f2ef. Added a `queueJobEnqueued` flag that flips to `true` immediately after `queue.add` succeeds. The `markMessageFailed` call in the catch is now guarded by `!queueJobEnqueued`, so only pre-enqueue failures mark the message FAILED. Post-enqueue writes (`updateConversationState`, `notifyStateUpdated`) are wrapped in their own best-effort try/catch so any transient failure there cannot reach the outer catch and poison a live job's state.

Gautam2086:
Addressed Greptile's two findings:

**Issue 2 (import path):** Fixed in c093849 — both `markMessageFailed` dynamic imports in the worker now use the shorter path `../../chat/tools`.

**Issue 1 (unreachable catch in start.ts):** Confirmed pre-existing dead code — `runDeepResearch` is `async` so the surrounding `try/catch` can never fire synchronously. The `markMessageFailed` we added there is equally unreachable. Greptile correctly notes the real coverage comes from `handleDeepResearchStartFailure` inside `runDeepResearch`. Cleaning up the dead block is a follow-up since it's outside the scope of this fix.

Gautam2086:

raghav-rama (approved):
Looks good. Merge if tested.

PR #183, opened 2026-05-27
feat(tools): binder target pipeline proxy routes (stages 1-3)

Gautam2086:
## Summary

This PR covers the BioAgents backend for the binder target pipeline (stages 1-3), including both the proxy routes (v1) and the target chat-tool integration (v2).

### v1 - Proxy routes

Converts `src/routes/tools/target.ts` and `src/routes/tools/contacts.ts` from ~950-line v0 TypeScript implementations to thin proxy routes forwarding to bio-literature-api, where all business logic lives.

- `POST /api/tools/target` - proxies to `BIO_LIT_AGENT_API_URL/tools/target`
- `GET /api/tools/target/contacts` - proxies to `BIO_LIT_AGENT_API_URL/tools/target/contacts`
- `POST /api/tools/target/p2rank` - proxies to `BIO_LIT_AGENT_API_URL/tools/target/p2rank` (120s timeout for Modal cold start)
- `GET /api/tools/pdb-proxy` - kept in BioAgents (high-frequency 3D viewer calls, no business logic)
- `GET /api/tools/alphafold/proxy` - kept in BioAgents (same reason)
- Adds `"tools"` rate-limit bucket (30 req/min) to `rateLimiter.ts`
- Deletes `src/utils/alignment.ts` (NW alignment ported to Python in bio-literature-api)

Proxy implementation notes:
- `set.status = res.status` on all proxied responses - 404/500 from bio-lit-api forwarded exactly, never swallowed
- `AbortController` + 60s timeout on target/contacts, 120s on p2rank (Modal cold start)
- GET contacts proxy reconstructs query string via `url.searchParams.set`, not forwarded as POST body
- Uses existing `BIO_LIT_AGENT_API_URL` and `BIO_LIT_AGENT_API_KEY` env vars

### v2 - Target chat-tool pipeline

Wires `/target` slash command into the persisted chat artifact pipeline, following the same pattern as the existing Segment Anything integration.

- `src/services/target/chat-tool.ts` - calls bio-lit-api `/tools/target`, returns `DataArtifact` with type `target-result`; full `TargetResponse` stored in `artifact.metadata` alongside `_query` and `_version: 1` for schema tracking
- Extended `DataArtifact.type`, `ChatToolId`, and `TargetToolInput` in `src/types/core.ts`
- `parseChatToolId` in SAM chat-tool now also accepts `"target"`
- Target path wired in all three transports:
  - `src/routes/chat.ts` (non-SSE JSON path)
  - `src/routes/chat-sse-transport.ts` (SSE path - the normal chat path)
  - `src/services/queue/workers/chat.worker.ts` (BullMQ queue worker, production `USE_JOB_QUEUE=true`)
- Results stored via `persistNormalChatArtifacts` - persist across page refresh

### Tests

- 6 unit tests for `runTargetChatTool`: happy path artifact shape, metadata fields, message-as-query fallback, 503 (missing env), 502 (upstream error), 400 (empty query)
- 2 SSE transport tests mirroring the SAM cases: artifact persists before final event, DB failure routes to error event

## Test plan

- `bun typecheck && bun test` passes (284/284, 7 skip) - verified locally
- `POST /api/tools/target` with valid auth returns proxied result (uniprotId present)
- `POST /api/tools/target` unauthenticated returns 401
- 404 from bio-lit-api (invalid gene) forwarded as 404, not 200 or 502
- `/target GLP1R` in staging chat renders TargetResult card inline, result persists on page refresh


src/routes/tools/contacts.ts
```diff
@@ -0,0 +1,36 @@
...
+        for (const [k, v] of Object.entries(query as Record<string, string>)) {
+          if (v !== undefined) url.searchParams.set(k, v);
+        }
+        const res = await fetch(url.toString(), {
+          headers: { "X-API-Key": BIO_LIT_KEY() ?? "" },
+          signal: controller.signal,
+        });
+        set.status = res.status;
+        return await res.json();
+      } catch {
```
jiggyjo11: **Critical — bare `catch {}` swallows all upstream context.**

Timeouts (`AbortError`), JSON parse failures, DNS errors, and auth rejections all collapse into the same generic `"Target contacts upstream error"` 502 with **zero logging**. Operators can't tell a 60s timeout from a malformed response.

Fix: bind the error and log it (matches `alphafoldProxy.ts:42`):
```ts
} catch (err) {
  logger.warn({ err, url: BIO_LIT_URL() }, "target_contacts_proxy_failed");
  set.status = 502;
  return { error: "Target contacts upstream error" };
}
```

src/routes/tools/contacts.ts
```diff
@@ -0,0 +1,36 @@
...
+        const url = new URL(`${BIO_LIT_URL()}/tools/target/contacts`);
+        for (const [k, v] of Object.entries(query as Record<string, string>)) {
+          if (v !== undefined) url.searchParams.set(k, v);
+        }
+        const res = await fetch(url.toString(), {
+          headers: { "X-API-Key": BIO_LIT_KEY() ?? "" },
+          signal: controller.signal,
+        });
+        set.status = res.status;
+        return await res.json();
```
jiggyjo11: **Critical — `await res.json()` inside the same `try` mis-classifies non-JSON upstream errors.**

When upstream returns HTML (e.g. nginx 502 page, gateway timeout body), this throws a `SyntaxError`, falls into the bare `catch`, and the real upstream status code (already assigned to `set.status` on the line above) is silently discarded — client sees a generic 502.

Guard with content-type or wrap the parse in its own try:
```ts
set.status = res.status;
const ct = res.headers.get("content-type") ?? "";
if (!ct.includes("application/json")) {
  const text = await res.text().catch(() => "");
  return { error: `Non-JSON upstream response`, status: res.status, body: text.slice(0, 500) };
}
return await res.json();
```

src/routes/tools/contacts.ts
```diff
@@ -0,0 +1,36 @@
...
+import { rateLimitMiddleware } from "../../middleware/rateLimiter";
+
+const BIO_LIT_URL = () => process.env.BIO_LIT_AGENT_API_URL;
+const BIO_LIT_KEY = () => process.env.BIO_LIT_AGENT_API_KEY;
+
+export const contactsRoute = new Elysia().guard(
+  { beforeHandle: [authResolver({ required: true }), rateLimitMiddleware("tools")] },
+  (app) =>
+    app.get("/api/tools/target/contacts", async ({ query, set }) => {
+      if (!BIO_LIT_URL()) {
```
jiggyjo11: **Critical — missing `BIO_LIT_AGENT_API_KEY` guard.**

Only the URL is checked. When the key is missing/empty, the request goes upstream with `X-API-Key: ""` and the 401/403 is then surfaced as a confusing 502 (via the catch below). The service layer at `src/services/target/chat-tool.ts:40` gets this right — replicate the pattern:

```ts
if (!BIO_LIT_URL() || !BIO_LIT_KEY()) {
  set.status = 503;
  return { error: "Target service not configured" };
}
```

jiggyjo11 (commented):
## PR Review — /target slash command + tool proxies

> **Update (2026-06-01):** annotated below with status after commits `44e3264`, `5234dc1`, `7ed4e97`, `5357a3e`. Follow-up review with new findings posted as a separate comment on this PR.

Scope: 17 files, ~900 LoC since merge-base `8517306` (target chat-tool + SSE/queue wiring + 5 proxy routes + rate-limit bucket).

### Critical (must fix before merge)

1. ⚠️ **Partially open** — **Bare `catch {}` in 3 proxy routes** swallows all upstream context — timeouts, JSON parse failures, auth rejections, network errors all collapse into the same generic 502 with **zero logging**. `alphafoldProxy.ts:42` / `pdbProxy.ts` already show the right pattern. _The catch was kept but is still anonymous and silent on the JSON-parse path; see follow-up review._
2. ⚠️ **Still open** — **`await res.json()` inside the same `try` mis-classifies non-JSON upstream errors** — when upstream returns HTML (nginx 502) the parse throws, falls into the bare catch, and the real upstream status code (already on `set.status`) is discarded. _Same shape at `target.ts:40-45`, `p2rank.ts:40-45`, `contacts.ts:39-44`._
3. ✅ **Fixed** in 5357a3e — **Missing `BIO_LIT_AGENT_API_KEY` guard** in the 3 BIO_LIT proxy routes — requests go upstream with `X-API-Key: ""` and 401/403 surfaces as a confusing 502. The service layer (`chat-tool.ts:40`) gets this right. _Now all 3 proxies check `BIO_LIT_KEY()` and return 503._
4. ✅ **Fixed** in 5357a3e — **`runTargetChatTool` doesn't wrap `await response.json()`** at `chat-tool.ts:74` — a 200 with non-JSON throws a raw `SyntaxError`, escapes unwrapped, and SSE/queue/in-process callers all fall through their `instanceof TargetChatToolError` branch — user sees `"Unexpected token < ..."`. _Now wrapped (`chat-tool.ts:75-79`), throws `TargetChatToolError("...non-JSON response", 502)`._

### Important

5. ⚠️ **Partially fixed** in 7ed4e97 — Queue-worker `runTargetForJob` (`chat.worker.ts:259-325`) and in-process `/target` branch (`chat.ts:728-791`) have **no tests** — only the SSE path got coverage. Queue is the production execution mode. _SSE transport + target chat-tool service now have solid tests, but `runTargetForJob` and the in-process `/target` branch in `chat.ts` are still uncovered, and so are all 5 proxy routes._
6. ✅ **Fixed** in 44e3264 — **`uniprotId` silent fallback to raw query** at `chat-tool.ts:89` — when upstream's `target.uniprotId` is missing/non-string, the user-facing text becomes `"Target analysis complete for kinase inhibitor"` (their raw query) as if everything worked. _Root cause was reading `targetData.uniprotId` (always undefined) instead of `targetData.target.uniprotId`. Now reads the nested field correctly. (The `?? input.query` fallback remains, but is now a true cold-path edge case rather than the hot path.)_
7. ✅ **Fixed** in 5357a3e — **`alphafoldProxy.ts:18` allowlist uses `startsWith`** — vulnerable to `@user:host` / path tricks; use `new URL(url).host === "alphafold.ebi.ac.uk"`. Also `alphafoldProxy` + `pdbProxy` have no rate limiter; add `rateLimitMiddleware("tools")`. _Now `parsedUrl.host !== "alphafold.ebi.ac.uk" || parsedUrl.protocol !== "https:"` (`alphafoldProxy.ts:26`); rate limiter wired on all 5 proxies._
8. ✅ **Fixed** in 5357a3e — **`markMessageFailed` not guarded** at `chat.ts:740` — DB failure shadows the original `TargetChatToolError` with a DB stack. _Now `await markMessageFailed(...).catch((dbErr) => logger.error(...))` at `chat.ts:740-742`._
9. ❌ **Still open** — **Upstream response not Zod-validated** — `chat-tool.ts:74` blind-casts to `Record<string, unknown>`. `src/types/target.ts` is 93 lines of advisory `interface`s with zero runtime enforcement. Highest-ROI followup.

### Suggestions

10. ❌ **Still open** — **Duplicated proxy boilerplate** — `target.ts`, `p2rank.ts`, `contacts.ts` are ~95% identical; extract `createBioLitProxy({ path, method, timeoutMs })`. Same for the two CORS proxies.
11. ❌ **Still open** — **`TargetChatToolError.statusCode`** should be `readonly`.
12. ❌ **Still open** — **`DataArtifact`** isn't a discriminated union over `type` — adding `"target-result"` without payload typing leaves `metadata: Record<string, unknown>` as an escape hatch.
13. ❌ **Still open** — **Rate-limit bucket `tools`** untested for default + env-override parsing.
14. ✅ **Fixed** in 5357a3e — **`logger.warn` for production upstream failures** (`chat-tool.ts:59, 67`) — repo convention is `logger.error` with an `errorId` so Sentry catches them. _Both call sites are `logger.error` now._
15. ❌ **Still open** — **Inconsistent timeouts** across proxies (15s / 60s / 65s / 120s) with no shared constants — drift risk.

### Strengths

- SSE wiring for `target` is a faithful structural mirror of the proven `segment-anything` path: same `replyPersisted` flag, same persist → markComplete → notify ordering, same error→failed downgrade guard.
- Worker path uses `failJobIfRowNoLongerPending` and tolerates notify errors — won't dead-end retries.
- Unit coverage of `runTargetChatTool` is strong (6 tests). `target/__tests__/chat-tool.test.ts:96-101` correctly asserts upstream body text doesn't leak in thrown messages.
- `chat-sse-transport.test.ts:354-384` covers the persistence-failure-after-tool-success rollback — exactly the kind of multi-step bug that's easy to regress.

_Review generated via `/pr-review-toolkit:review-pr` (code-reviewer + pr-test-analyzer + silent-failure-hunter + type-design-analyzer)._

src/routes/tools/p2rank.ts
```diff
@@ -0,0 +1,37 @@
...
+          headers: {
+            "Content-Type": "application/json",
+            "X-API-Key": BIO_LIT_KEY() ?? "",
+          },
+          method: "POST",
+          signal: controller.signal,
+        });
+        set.status = res.status;
+        return await res.json();
+      } catch {
```
jiggyjo11: **Critical — bare `catch {}` swallows upstream errors.** Same issue as `contacts.ts:29` and `target.ts:30`. No logging means timeouts, parse failures, and auth rejections are indistinguishable from generic 502s. Bind and log `err` like `alphafoldProxy.ts:42`.

src/routes/tools/p2rank.ts
```diff
@@ -0,0 +1,37 @@
...
+          body: JSON.stringify(body),
+          headers: {
+            "Content-Type": "application/json",
+            "X-API-Key": BIO_LIT_KEY() ?? "",
+          },
+          method: "POST",
+          signal: controller.signal,
+        });
+        set.status = res.status;
+        return await res.json();
```
jiggyjo11: **Critical — `await res.json()` inside the same `try` mis-classifies non-JSON 4xx/5xx upstream responses.** See `contacts.ts:28` for the full pattern + fix.

src/routes/tools/p2rank.ts
```diff
@@ -0,0 +1,37 @@
...
+import { rateLimitMiddleware } from "../../middleware/rateLimiter";
+
+const BIO_LIT_URL = () => process.env.BIO_LIT_AGENT_API_URL;
+const BIO_LIT_KEY = () => process.env.BIO_LIT_AGENT_API_KEY;
+
+export const p2rankRoute = new Elysia().guard(
+  { beforeHandle: [authResolver({ required: true }), rateLimitMiddleware("tools")] },
+  (app) =>
+    app.post("/api/tools/target/p2rank", async ({ body, set }) => {
+      if (!BIO_LIT_URL()) {
```
jiggyjo11: **Critical — missing `BIO_LIT_AGENT_API_KEY` guard.** Only URL is checked; missing key gives confusing 502s. See `contacts.ts:12` for the fix pattern.

src/routes/tools/target.ts
```diff
@@ -0,0 +1,37 @@
...
+          headers: {
+            "Content-Type": "application/json",
+            "X-API-Key": BIO_LIT_KEY() ?? "",
+          },
+          method: "POST",
+          signal: controller.signal,
+        });
+        set.status = res.status;
+        return await res.json();
+      } catch {
```
jiggyjo11: **Critical — bare `catch {}` swallows upstream errors.** Same as `contacts.ts:29` / `p2rank.ts:30`. Bind and log `err`.

src/routes/tools/target.ts
```diff
@@ -0,0 +1,37 @@
...
+          body: JSON.stringify(body),
+          headers: {
+            "Content-Type": "application/json",
+            "X-API-Key": BIO_LIT_KEY() ?? "",
+          },
+          method: "POST",
+          signal: controller.signal,
+        });
+        set.status = res.status;
+        return await res.json();
```
jiggyjo11: **Critical — `await res.json()` inside the same `try` mis-classifies non-JSON upstream errors.** See `contacts.ts:28` for fix.

src/routes/tools/target.ts
```diff
@@ -0,0 +1,37 @@
...
+import { rateLimitMiddleware } from "../../middleware/rateLimiter";
+
+const BIO_LIT_URL = () => process.env.BIO_LIT_AGENT_API_URL;
+const BIO_LIT_KEY = () => process.env.BIO_LIT_AGENT_API_KEY;
+
+export const targetRoute = new Elysia().guard(
+  { beforeHandle: [authResolver({ required: true }), rateLimitMiddleware("tools")] },
+  (app) =>
+    app.post("/api/tools/target", async ({ body, set }) => {
+      if (!BIO_LIT_URL()) {
```
jiggyjo11: **Critical — missing `BIO_LIT_AGENT_API_KEY` guard.** Only the URL is checked. See `contacts.ts:12`.

src/services/target/chat-tool.ts
```diff
@@ -0,0 +1,93 @@
...
+  if (!response.ok) {
+    const errorText = await response.text().catch(() => "");
+    logger.warn({ errorText, status: response.status }, "target_chat_tool_upstream_error");
+    throw new TargetChatToolError(
+      `Target pipeline error: ${response.status}`,
+      response.status >= 500 ? 502 : response.status
+    );
+  }
+
+  const targetData = (await response.json()) as Record<string, unknown>;
```
jiggyjo11: **Critical — unwrapped `await response.json()` throws a raw `SyntaxError`.**

If upstream returns 200 with a non-JSON body (corruption, truncation, gateway HTML), this throws a `SyntaxError` that escapes the function unwrapped. All three callers (`chat.ts:736`, `chat.worker.ts:266`, `chat-sse-transport.ts:367-ish`) check `instanceof TargetChatToolError` and fall through to their generic-error branch — user sees `"Unexpected token < in JSON at position 0"`.

Wrap and reclassify as `TargetChatToolError("Invalid response from target pipeline", 502)`.

**Important — also no Zod validation of the response shape.** The cast to `Record<string, unknown>` makes the 93-line `src/types/target.ts` purely advisory. Highest-ROI followup: define a Zod schema for `TargetData` and `safeParse` here, throwing `TargetChatToolError("Malformed target pipeline response", 502)` on mismatch.

src/services/target/chat-tool.ts
```diff
@@ -0,0 +1,93 @@
...
+      ...targetData,
+      _query: input.query,
+      _version: 1,
+    },
+    name: `Target: ${input.query}`,
+    type: "target-result",
+  };
+
+  const targetInfo = targetData.target as Record<string, unknown> | undefined;
+  const uniprotId = typeof targetInfo?.uniprotId === "string" ? targetInfo.uniprotId : input.query;
```
jiggyjo11, line 120: **Important — silent fallback to the user's raw query is a UX bug.**

When upstream's `target.uniprotId` is absent or non-string (malformed response), the text becomes `"Target analysis complete for kinase inhibitor"` (their free-form query) — looking like a successful UniProt lookup. The malformed payload is never logged either.

Either:
- log `logger.warn({ targetInfo }, "target_chat_tool_missing_uniprot_id")` and use a clearer message (`"Target analysis complete (no UniProt ID returned)"`), or
- treat the missing ID as a hard error (throw `TargetChatToolError("Target pipeline did not return a uniprotId", 502)`).

A branded `UniProtId` type with the canonical accession regex would lock this down structurally.

src/services/target/chat-tool.ts
```diff
@@ -0,0 +1,93 @@
+import type { DataArtifact, TargetToolInput } from "../../types/core";
+import logger from "../../utils/logger";
+
+const TARGET_FETCH_TIMEOUT_MS = 65_000;
+
+export class TargetChatToolError extends Error {
+  statusCode: number;
```
jiggyjo11: **Suggestion — `statusCode` should be `readonly`.** Mutable error fields invite subtle bugs (a downstream handler reassigning the status after throw). Also consider narrowing the constructor type from `number` to the small set you actually emit (`400 | 502 | 503` plus passthrough).

src/services/target/chat-tool.ts
```diff
@@ -0,0 +1,93 @@
...
+      body: JSON.stringify({ query: input.query }),
+      headers: {
+        "Content-Type": "application/json",
+        "X-API-Key": apiKey,
+      },
+      method: "POST",
+      signal: controller.signal,
+    });
+  } catch (err) {
+    logger.warn({ err }, "target_chat_tool_request_failed");
```
jiggyjo11: **Suggestion — `logger.warn` for production upstream failures.** Per repo convention these go to Sentry as `logger.error` with an `errorId`. Same at line 67. As-is, real production failures will be lost in warn-level noise.

src/routes/tools/alphafoldProxy.ts
```diff
@@ -0,0 +1,50 @@
...
+    app.get("/api/tools/alphafold/proxy", async ({ query, set }) => {
+      const { url } = query as { url?: string };
+
+      if (!url) {
+        set.status = 400;
+        return { error: "Missing url parameter" };
+      }
+
+      // Only allow AlphaFold EBI URLs to prevent open proxy abuse
+      if (!url.startsWith("https://alphafold.ebi.ac.uk/")) {
```
jiggyjo11: **Important — allowlist using `startsWith` is bypassable.**

`startsWith("https://alphafold.ebi.ac.uk/")` accepts URLs like `https://alphafold.ebi.ac.uk/@evil.com/...` (userinfo + redirect) or with embedded `..`. Use structured parsing:

```ts
let target: URL;
try { target = new URL(url); } catch { set.status = 400; return { error: "Invalid url" }; }
if (target.protocol !== "https:" || target.host !== "alphafold.ebi.ac.uk") {
  set.status = 400; return { error: "Only AlphaFold EBI URLs are allowed" };
}
```

**Important — also no rate limiter.** This route + `pdbProxy.ts` are open authenticated proxies to external services. Add `rateLimitMiddleware("tools")` to the guard at line 7.

src/routes/chat.ts
```diff
@@ -721,6 +722,74 @@ export async function chatHandler(ctx: ElysiaRouteContext) {
...
+        targetResult = await runTargetChatTool({
+          message,
+          messageId: createdMessage.id,
+          toolInput,
+        });
+      } catch (err) {
+        const statusCode = err instanceof TargetChatToolError ? err.statusCode : 502;
+        const detail = err instanceof Error ? err.message : "Target pipeline failed";
+        logger.warn({ err, messageId: createdMessage.id }, "target_chat_tool_error");
+        await markMessageFailed(createdMessage.id);
```
jiggyjo11: **Important — `markMessageFailed` not guarded.**

If the DB write fails here, the original `TargetChatToolError` is shadowed by a DB error and the user sees a DB stack instead of the real failure reason. Wrap:

```ts
try { await markMessageFailed(createdMessage.id); }
catch (dbErr) { logger.error({ dbErr, origErr: err, messageId: createdMessage.id }, "target_mark_failed_db_error"); }
```

src/routes/chat.ts
```diff
@@ -721,6 +722,74 @@ export async function chatHandler(ctx: ElysiaRouteContext) {
...
+      );
+
+      return new Response(JSON.stringify(response), {
+        headers: {
+          "Content-Encoding": "identity",
+          "Content-Type": "application/json; charset=utf-8",
+        },
+        status: 200,
+      });
+    }
```
jiggyjo11, line 793: **Important — in-process `/target` branch has no test coverage.**

Only the SSE transport (`src/routes/chat-sse-transport.test.ts:321-383`) and the unit-level service got tests. This entire in-process branch (the default path when `Accept: text/event-stream` isn't set) is untested. Worth asserting:
- `TargetChatToolError(503)` → 503 HTTP status
- `markMessageFailed` invoked on error
- `markMessageComplete` not-updated branch returns 500

src/services/queue/workers/chat.worker.ts
```diff
@@ -243,6 +255,76 @@ async function runSegmentAnythingForJob(
...
+      "chat_worker_target_message_updated_notify_failed"
+    );
+  }
+
+  return {
+    artifacts: targetResult.artifacts,
+    responseTime,
+    text: targetResult.text,
+    userId,
+  };
```
jiggyjo11, line 330: **Important — `runTargetForJob` has no tests.**

This is the production execution mode (queue mode). The SSE transport got two tests for the target path, the queue worker got none. A regression in `notifyJobCompleted` payload shape, error rethrow behavior, or the `failJobIfRowNoLongerPending` branch will silently pass CI.

Mirror the SSE tests against this worker.

src/types/core.ts
```diff
@@ -88,7 +88,7 @@ export type ProteinStructure = {
 export type DataArtifact = {
   id: string;
   name: string;
-  type: "table" | "chart" | "notebook" | "code" | "file" | "image" | "pdf";
+  type: "table" | "chart" | "notebook" | "code" | "file" | "image" | "pdf" | "target-result";
```
jiggyjo11, line 91: **Suggestion — `DataArtifact` is no longer a flat type once `"target-result"` is added.**

The union now carries an implicit invariant (`type: "target-result"` ⇒ `metadata: TargetData`-shaped) that's invisible to the type system. `metadata: Record<string, unknown>` is an escape hatch — consumers cannot get typed access to the target data without re-narrowing.

Consider a discriminated union:
```ts
type DataArtifactBase = { id: string; name: string; description?: string; /* ... */ };
type TargetResultArtifact = DataArtifactBase & { type: "target-result"; metadata: TargetData };
type GenericArtifact = DataArtifactBase & { type: Exclude<...>; metadata?: Record<string, unknown> };
export type DataArtifact = TargetResultArtifact | GenericArtifact;
```

jiggyjo11 (commented):
## Follow-up Review — full multi-agent pass on `ga-binder-stages-proxy` @ 5357a3e

Scope: 21 files vs `origin/dev` (~750 net LoC). Runs after the previous review fixes landed in 5357a3e, 7ed4e97, 5234dc1, 44e3264. Five specialised agents ran (code-reviewer, silent-failure-hunter, type-design-analyzer, pr-test-analyzer, comment-analyzer); findings deduplicated below. **Previous review body has been edited in place with fixed/open status for each item.**

The proxy/tool work itself is in good shape after the round-trip. The blocker is a merge hazard, not a content issue.

---

### 🚨 Critical (block merge)

**1. Branch is missing `d32e53e` (BIOS-94) — merging will silently revert the deep-research failure-handling fix.**
`d32e53e` ("fix: DR failure paths now flip message row to FAILED") is on `origin/dev` but not on this branch. Verified: HEAD's `src/routes/deep-research/start.ts` and `src/services/queue/workers/deep-research.worker.ts` contain **zero** `markMessageFailed` calls; `origin/dev`'s versions reintroduced them as the BIOS-94 fix. The PR diff against `origin/dev` correspondingly shows `src/routes/deep-research/failure-handler.ts` being recreated (144 lines), `src/routes/deep-research/__tests__/start.test.ts` being re-deleted (98 lines), and 191/113-line "reverts" of `start.ts`/`deep-research.worker.ts`. None of that is in this PR's intended scope ("stages 1-3 proxy"). On merge, deep-research failures will leave the message row PENDING indefinitely.
**Fix:** rebase `ga-binder-stages-proxy` onto `origin/dev`, re-run `bun typecheck && bun test`. The 4 deep-research files should drop out of the diff entirely. Do not merge until that diff is clean.

---

### ⚠️ Important (fix before merge)

**2. Three proxies still leak upstream error bodies + silently swallow JSON-parse failures.** This is the unresolved tail of items 1+2 in the prior review. `src/routes/tools/{target,p2rank,contacts}.ts:38-44` all do:
```ts
set.status = res.status;
try { return await res.json(); } catch { return { error: `... returned ${res.status}` }; }
```
- **Leaks**: upstream stack traces / internal hostnames are forwarded verbatim on 4xx/5xx. `services/target/chat-tool.ts:55-62` correctly sanitises — proxies should match.
- **Silent**: anonymous catch swallows the parse error with no log; if upstream returns `200 OK` with a malformed body, client gets `200` + `{error: "...returned 200"}` — contradictory payload.
- **Logging asymmetry**: `pdbProxy.ts` and `alphafoldProxy.ts` log upstream non-OK and translate to 502; these three don't.
**Fix:** check `!res.ok` first, log on parse failure, return `502` on parse failure regardless of upstream status. Also apply the `bioLitUrl.replace(/\/$/, "")` trailing-slash normalisation that `chat-tool.ts:37` uses (`new URL(\`${bioLitUrl}/tools/...\`)` will currently produce `//tools/...` on a trailing-slash env value).

**3. Proxy POST bodies aren't validated.** `target.ts:23` / `p2rank.ts:23` forward `JSON.stringify(body)` straight to upstream — no Elysia `t.Object` schema, no shape check. `body` is whatever Elysia parsed; if a client posts no JSON body the request goes upstream as the literal `undefined`. More importantly, an authenticated client can shovel arbitrary nested attacker-controlled payload through the proxy. Compare `runTargetChatTool` (`chat-tool.ts:18-26`) which validates `{query: string, min 1}` itself.
**Fix:** add Elysia body schemas matching `target` / `p2rank` upstream contracts. Same applies to typing `contacts` query params (currently `query as Record<string, string>`).

**4. Worker retries non-recoverable target errors.** `chat.worker.ts:275-279` throws `TargetChatToolError` back to BullMQ, which retries up to 3× even for 400 (empty query) or 503 (missing env). The chat-agent path at `chat.worker.ts:349-352` correctly uses `UnrecoverableError` for misconfig.
**Fix:** wrap `<500` statuses in `UnrecoverableError` like the chat-agent path does.

**5. `TargetChatToolError` not handled in `chat.ts` outer catch.** `chat.ts:935-941` only guards `SegmentAnythingToolError`. If anything after the inner try/catch throws (e.g. `persistNormalChatArtifacts`, `markMessageComplete`), the outer catch leaks `err.message` and returns 500 instead of the right status — also inconsistent with the SA hide-5xx-detail policy.
**Fix:** extract a `ChatToolError` interface (`statusCode`, `clientMessage`) and use one `instanceof` check, or add `TargetChatToolError` to the existing guard symmetrically.

**6. Test coverage gap on every new proxy route + worker target branch.** Five new public endpoints; only the target *chat-tool service* and SSE transport are tested. None of `tools/{target,p2rank,contacts,pdbProxy,alphafoldProxy}.ts` have tests. Highest-risk untested code: **the alphafoldProxy SSRF guard** (`alphafoldProxy.ts:18-29`) — the comment at line 18 explicitly calls out the userinfo bypass, but nothing pins that behaviour. The `runTargetForJob` branch in `chat.worker.ts:255-325` is also still untested (item 5 from prior review).
**Fix punch list (ordered by risk):**
- alphafoldProxy SSRF — reject cases for userinfo bypass (`https://a@evil.com/`), host suffix (`alphafold.ebi.ac.uk.evil.com`), wrong scheme, non-URL strings; assert `fetch` not called.
- pdbProxy regex (`^[A-Za-z0-9]{4}$`), uppercase normalisation, RCSB 404 → 404 vs other → 502 mapping.
- target/p2rank/contacts: env-missing → 503, AbortError → 504, other fetch error → 502, malformed JSON.
- `runTargetForJob` happy path + `TargetChatToolError` propagation.

**7. New `tools` rate-limit bucket is untested.** A typo (`"tool"` vs `"tools"`) in `rateLimiter.ts:36-40` would silently disable limiting on all five proxies. Add at least one 429 assertion.

---

### 💡 Suggestions

**8. `src/types/target.ts` is dead.** `TargetData` and the nine helper interfaces are never imported; `runTargetChatTool` annotates the upstream response as `Record<string, unknown>` and stuffs it into `metadata`. Either (a) make the types load-bearing by adding a Zod schema, deriving `TargetData = z.infer<...>`, and parsing the upstream response at the proxy boundary — which would also catch silent BioLit schema drift — or (b) delete the file. Currently it pays no rent. (This is the same concern as item 9 in the prior review, scoped to where the fix has highest leverage.)

**9. Illegal states representable in `TargetData`.** `contactPdbId`/`contactChain` are independently nullable but co-locked; same for the `gpcrdbEntry`/`gpcrPocketResidues`/`gpcrSegments` trio. Source-tag fields (`UniProtFeature.type`, `*.source`, `annotationSources[]`) are open `string` despite small closed vocabularies. If you adopt the Zod-at-boundary approach in #8, modelling these as discriminated unions becomes a one-line addition. Quantitative ratings: encapsulation 2/5, invariant expression 1/5, usefulness 2/5, enforcement 1/5.

**10. Cache-Control on auth-gated proxies.** `alphafoldProxy.ts:46` and `pdbProxy.ts:36` set `Cache-Control: public, max-age=86400` on responses that require auth. Payloads are public assets so `public` is semantically defensible, but a shared CDN that doesn't key on `Authorization` could serve cached copies to other users. Consider `private` unless the CDN behaviour is verified.

**11. Comments to trim** (per CLAUDE.md "default to no comments"):
- `chat.ts:725-727` — 3-line `=====` banner for one `if (toolId === "target")` block.
- `index.ts:259-263` — 5 inline trailing comments restating routes' HTTP verb + path.
- `pdbProxy.ts:6`, `alphafoldProxy.ts:6` — restate filename + hardcoded host.
- Keep `alphafoldProxy.ts:18` — the one added comment that earns its place (security WHY for URL-parse over `.startsWith()`).

**12. SSE transport `controller.enqueue` catches are too broad.** `chat-sse-transport.ts:129-133, 142-148, 163-170` use anonymous `catch {}` to absorb client-disconnect, but will also swallow `JSON.stringify` circular-ref errors and encoder errors. Narrow to `TypeError`/"controller closed" and log other errors.

**13. Discoveries can be silently lost.** `start.ts:1146-1151` and `deep-research.worker.ts:423-432` `state_write_chain_error_suppressed` correctly prevents a poisoned chain, but the user never learns their research state is partial. Set `conversationState.values.deepResearchRun.hasWriteFailures = true` so it surfaces. (Note: these files should also disappear from the diff after the #1 rebase.)

**14. Credit-refund failures are logged but not actionable.** `deep-research.worker.ts:807-832` — billing-critical RPC failures should enqueue a reconciliation job or hit a high-priority error id, not just `logger.error`. (Same note re: rebase.)

---

### ✅ Strengths

- **SSRF guard via URL parsing** (`alphafoldProxy.ts:18-29`) correctly defeats userinfo-bypass and host-suffix attacks — exactly the right primitive, with a comment that explains *why*. Direct fix for item 7 in the prior review.
- **PDB ID regex validation** before path interpolation (`pdbProxy.ts:13-16`).
- **Env vars read inside functions** in the new chat-tool service — TDZ-safe per CLAUDE.md.
- **`AbortController` + timeouts** on all five proxies; differentiated per-route.
- **`runTargetChatTool` sanitises upstream error text** in the thrown error while still logging it server-side — defence in depth that the proxy routes should now copy.
- **`runTargetForJob` mirrors `runSegmentAnythingForJob`** structure cleanly; new rate-limit bucket is cleanly partitioned.
- **`chat-sse-transport.test.ts`** is exemplary: DAMP names, ordering assertions, regression annotations referencing PR #179.
- **Idempotent `replyPersisted` guard** verified preventing FAILED downgrade after persistence.
- **Docker-compose** wires `BIO_LIT_AGENT_API_URL`/`_API_KEY` to the API service correctly.

---

### Recommended action plan

1. **Rebase onto `origin/dev`** — drops 4 deep-research files from the diff, undoes the BIOS-94 revert. Re-run `bun typecheck && bun test`. **Don't merge before this step.**
2. Fix proxy error handling + sanitisation (#2) and add body validation (#3) — small, contained, copy the chat-tool service's hygiene.
3. Add the `UnrecoverableError` wrap (#4) and unify chat-tool error classes (#5).
4. Add the alphafold SSRF tests + a 429 test for the new bucket (#6, #7) — minimum to ship safely.
5. Defer the Zod-at-boundary refactor (#8/#9) to a follow-up unless you're already in the file.

_Review generated via `/pr-review-toolkit:review-pr` (code-reviewer + pr-test-analyzer + silent-failure-hunter + type-design-analyzer + comment-analyzer)._

Gautam2086:
Hey Johannes (@jiggyjo11), just pushed 3f22647 which closes out all seven of the must-fix items from this review.

The rebase onto origin/dev is done, BIOS-94 is back in, and the four deep-research files dropped out of the diff cleanly. The proxy error handling got a full overhaul on all three BioLit routes: they now split on ok/not-ok before touching the response body. A 5xx returns 502 with a generic message so no upstream stacktrace or hostname ever leaks. A 4xx forwards the status and body since that context is genuinely useful on the client side. Non-JSON on a 2xx path is logged and returns 502 with a clear message. Trailing slash normalization is in on BIO_LIT_AGENT_API_URL too.

Body validation is wired via t.Object schemas on all three routes so Elysia rejects at 422 before the handler even runs. TargetChatToolError is now in the outer chatHandler catch alongside SegmentAnythingToolError, with the same 5xx message normalization applied in both the SSE inner catch and the outer handler. The worker wraps 4xx and 503 in UnrecoverableError so BullMQ stops retrying those.

Test count went from 8 to 28 in proxy-routes.test.ts, covering SSRF cases, validation errors, 5xx leak prevention, 404 forwarding, trailing slash behavior, and a mock.module based 429 assertion for the tools bucket. The runTargetForJob BullMQ branch is documented as a known gap via test.skip since properly testing it would need mocking of Job, markMessageComplete, persistNormalChatArtifacts, and all the notify calls, which would be pretty brittle scaffolding for a thin orchestration wrapper.

On the suggestions: the Zod schema at the proxy boundary and the DataArtifact discriminated union are both solid ideas that I want to pick up in a follow-up rather than expanding scope here. Same for the SSE transport catch narrowing. (statusCode readonly was addressed in b61b69a.) The Cache-Control note on the two CORS proxies is worth a separate conversation since it depends on CDN keying behavior we would need to verify. Items 13 and 14 touch deep-research files so I will leave those for whoever is next in those.

Good for another look when you get a chance.

src/routes/tools/contacts.ts
Gautam2086: Fixed in af1d3fe and 3f22647 across all three proxy routes. Replaced the bare catch with typed AbortError detection for the fetch failure, and moved res.json() into its own try block at the end so fetch errors and JSON parse failures stay on separate code paths with the right status codes.

src/routes/tools/contacts.ts
Gautam2086: Fixed in the same commits as the bare catch. res.json() now lives in a separate outer try block after the res.ok check, so a parse failure on a 2xx response returns 502 with a clear message rather than overwriting whatever status the upstream sent.

src/routes/tools/contacts.ts
Gautam2086: Fixed in af1d3fe. Both bioLitUrl and bioLitKey are now checked together upfront. Missing key returns 503 early the same way a missing URL does.

src/routes/tools/p2rank.ts
Gautam2086: Fixed in af1d3fe and 3f22647, same treatment as contacts.ts and target.ts. AbortError gets 504, other fetch errors get 502, and json parsing is in its own try block.

src/routes/tools/p2rank.ts
Gautam2086: Fixed. res.json() is now in a separate outer try block in all three proxy routes.

src/routes/tools/p2rank.ts
Gautam2086: Fixed in af1d3fe. Both URL and key are checked together, 503 early return if either is missing.

src/routes/tools/target.ts
Gautam2086: Fixed in af1d3fe and 3f22647. Same pattern as the other two routes: AbortError detection in the fetch catch block, json parse in its own separate try at the end.

src/routes/tools/target.ts
Gautam2086: Fixed. res.json() is now in a separate outer try block across all three proxy routes, same as the others.

src/routes/tools/target.ts
Gautam2086: Fixed in af1d3fe. Both bioLitUrl and bioLitKey checked together, 503 if either is missing.

src/services/target/chat-tool.ts
Gautam2086: Fixed in 3f22647. response.json() is now in its own try/catch that throws TargetChatToolError with status 502 on parse failure, so a malformed 2xx response doesn't surface as an unhandled SyntaxError.

src/services/target/chat-tool.ts
Gautam2086: Fixed in 3f22647. normalizeToolInput now validates the query strictly and throws a 400 TargetChatToolError if neither toolInput nor a non-empty message is provided. No more silent fallback to whatever string the user typed.

src/services/target/chat-tool.ts
Gautam2086: Good call. Fixed in b61b69a.

src/services/target/chat-tool.ts
Gautam2086: Fair point on the levels. Looking at the diff, it seems the code already has logger.warn for the request failure and upstream error paths (those were updated in 3f22647 alongside the other fixes). If anything slipped through to logger.error, let me know and I will clean it up.

src/routes/tools/alphafoldProxy.ts
Gautam2086: Fixed in 3f22647. The route now parses the URL with new URL() and checks parsedUrl.host === 'alphafold.ebi.ac.uk' plus parsedUrl.protocol === 'https:'. That closes the @user:host bypass the startsWith approach allowed.

src/routes/chat.ts
Gautam2086: Fixed in 3f22647. markMessageFailed is now wrapped in .catch() that logs the DB error separately, so a write failure doesn't mask the original TargetChatToolError or prevent the response from going out to the client.

src/routes/chat.ts
Gautam2086: Tests added in f943ab2. chat-sse-transport.test.ts covers the target SSE path specifically: one test for the happy path (artifacts only emitted after artifact state persists) and one for the failure case (persistence error sends an error event instead of artifacts).

src/services/queue/workers/chat.worker.ts
Gautam2086: Acknowledged as a gap. The worker path is documented with a test.skip in proxy-routes.test.ts explaining why it is integration-only: testing it properly would mean mocking BullMQ Job, markMessageComplete, persistNormalChatArtifacts, and all notify functions. A hollow unit test there would give false confidence. Integration coverage is the right call here.

src/types/core.ts
Gautam2086: Totally agree on the long-term direction. A proper discriminated union with per-type metadata shapes is the right move once there are more artifact types. Leaving it for a follow-up refactor so it doesn't block this PR, but I've noted it.

Gautam2086:

Gautam2086:

jiggyjo11 (approved):
LG!

Gautam2086:

